PR #5, opened 2016-12-16
Update Russian

mcbaze:
what about string "custom"? where is it used? didn't find anything about it..


tibbi:
hello, there are some strings not used anywhere, I just added them so they are translated by the time they go live. It will be used at the color customization screen, will be added soon.

PR #25, opened 2017-01-25
Improve spanish strings

Eskuero:
I left untouched release notes and Play Store description.
I'm only using the Music Player and not the Gallery one because the icon is too ugly lol
But if I find some time I'll try to translate it :)


tibbi:
heh thanks. The icon has been updated recently, but Im not fully satisfied with it either ;) Its better than it used to be tho.

Eskuero:
Is no that bad, probably I'm too used to CyanogenMod icons and switching to a complete family of orange ones feels weird haha, keep up the good work :)

PR #44, opened 2017-04-06
Update strings.xml

rodymacedo:
Upvated new strings. Not sure about the \ and n's placements, though. May need revision.


tibbi:
\n is a new line, you should use them in the same place and in the same amount as the original translation.
\' is escaping an apostrophe, so if there is no ' in the translated string, no back slash is needed.

PR #65, opened 2017-08-01
Ensure Context.getFileDocument reaches the desired path

Lonami:
Currently `Context.getFileDocument` may not reach the specified path. Say for instance the `/some/example.txt` file is trying to be accessed. But the directory `some/` doesn't exist, but `example.txt` does. Then, this file (`example.txt`) will be chosen instead the desired `/some/example.txt`, which may cause obscure data loss.

This fix ensures that the right path is chosen, even if it has to create it.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context-storage.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context-storage.kt
@@ -130,10 +130,24 @@ fun Context.getFileDocument(path: String): DocumentFile? {
 
     var document = DocumentFile.fromTreeUri(this, Uri.parse(baseConfig.treeUri))
     val parts = relativePath.split("/")
-    for (part in parts) {
-        val currDocument = document.findFile(part)
-        if (currDocument != null)
-            document = currDocument
+    for (i in 0..parts.size - 1) {
+        var currDocument = document.findFile(parts[i])
+        if (currDocument == null) {
+            // We need to assure that we transverse to the right directory!
+            if (i == parts.size - 1) {
+                // The last document should be the file we're looking for, not a directory
+                currDocument = document.createFile("", parts[i])
+            } else {
+                currDocument = document.createDirectory(parts[i])
+            }
+
+            if (currDocument == null) {
+                // Half paths should not be tolerated. If the given path cannot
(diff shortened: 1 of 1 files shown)


tibbi:
heh nice, Ill have to test it on a couple devices tho as the fileDocument is a bitch. If it fixed the issue some people have with deleted data, that would be great. Anyway if the document is not retrieved properly I dont want to crash the app, just let the user know through some toast that something failed. This might be happening on a background thread so not sure how to get a toast there, we will figure something out :)

Lonami:
> just let the user know through some toast that something failed

`catch` the exception. Blindly ignoring the errors is definitely not better than `throw`'ing exceptions.

> This might be happening on a background thread so not sure how to get a toast there

Probably some kind of callback (as I've seen somewhere else in the code) to let it know the operation was successful then. For my needs (single thread, for now…) this was the best solution I could think of.

By the way thanks for having a clear way to work with the file system on Lollipop and above ;) it really is a pain, I borrowed some code with a explicit "thanks to SimpleMobileTools for this bit".

tibbi:
I dont want to blindly ignore the errors, just rely on someone seeing the toast and writing me an email, then help debug :) It happened countless times, as there have been many issues which I couldnt reproduce on any device and some random people helped. Playstore crashdumps wouldnt really help me much, especially since the developer console has been reworked and I dont see the specific crash reason anymore. I see it like "java.lang.RuntimeException: ", but no detailed sentence after it, just the files and line numbers. Its happening with Java and proguard disabled too.

tibbi:
so it turns out that the getFileDocument was always called from activities, so I just moved it to the activity extensions and added a toast. It might create duplicate toasts in some scenarions, but its not a big deal at this point. Thanks for the patch, if it fixed the issue with wiping SD cards, that would be great :)

tibbi:
sadly I had to revert this as it was creating issues right at subfolders, many people couldnt delete/move files on SD cards. Ive added a different check for document filename, hopefully that will fix the sd card wiping.

Lonami:
Okay, thanks for the feedback and I hope your fix makes things better. I hope I didn't cause much trouble 😓

tibbi:
no problem, Ive fully reverted back to the old version as a modified one was causing issues too. I should start using Staged rollout properly already, so I dont spam everyone with new versions :)

PR #68, opened 2017-08-11
Updates and fixes

gregory678:


tibbi:
at the plurals, theres more versions available. There is for example the `<item quantity="one">%1$d sekunda</item>` with the "one" count, the second used count is "other". Theres however multiple other counts available, like "zero", "two", "few", "many". Feel free adding new lines with strings like `<item quantity="two">%1$d sekundy</item>` whenever needed. Thanks

gregory678:
OK, now I will know. I made a new version. This can be closed.

PR #69, opened 2017-08-11
Updates and fixes #2

gregory678:
Thanks for tips with the "plurals" thing.


tibbi:
looks better, thanks :) "few" means 2, 3, 4, so I need it at the slovak translation too.

PR #71, opened 2017-08-15
Update strings.xml

smarquespt:


commons/src/main/res/values-pt/strings.xml
```diff
@@ -183,7 +183,7 @@
     <string name="show_hidden">Mostrar ficheiros e pastas ocultos</string>
     <string name="font_size">Tamanho do texto</string>
     <string name="small">Pequeno</string>
-    <string name="medium">Medium</string>
+    <string name="medium">Médio</string>
     <string name="large">Grande</string>
     <string name="extra_large">Muito grande</string>
     <string name="password_protect_hidden_items">Password protect hidden item visibility</string>
```
smarquespt: I´m confused with this string. What exactly is this for?

tibbi: its used in the gallery and file manager. It adds password protection to toggling hidden album/folder visibility, so people cannot see it without knowing the password/pattern. Its just a simple protection from kids or so, no encryption.

smarquespt: So why not 'Protect hidden items with password'?

tibbi: that could also mean that the items are protected from deleting, not viewing.

PR #72, opened 2017-08-25
Added Korean translation

hanjoongcho:
Hi tibbi.
I added a Korean translation.
Have a nice day. =)


tibbi:
nice, thanks! But theres still some untranslated strings if you check https://github.com/SimpleMobileTools/Simple-Commons/blob/a4aec0b6eef45a79b20012c266736284b8940a54/commons/src/main/res/values-ko-rKR/strings.xml :p

PR #82, opened 2017-09-19
Update Japanese translation

mamizu:


tibbi:
nice, thanks for the translations

PR #107, opened 2017-11-10
Update strings.xml

gcalzo:
I didn't find "Audio permission" neither on Marshmallow nor on Nougat... do you mean "Mic permission"?


tibbi:
Well it is called Record Audio, guess I will rename it :)

gcalzo:
Ok, I will translate it to "Microfono" as is called into my Italian UI... :)

PR #108, opened 2017-11-10
Update strings.xml

gcalzo:


tibbi:
please escape quotes like this, I have to fix it manually to be able to compile the app. Ive fixed this one, Thanks

gcalzo:
Sorry, I forgot and since I've copy&paste from the first all were without... :(

PR #113, opened 2017-11-29
Updated Dutch

ltGuillaume:


tibbi:
@ltGuillaume I wrote a guide how to set up a file watcher which will send you an email whenever a file changes, you can check it out at https://github.com/SimpleMobileTools/General-Discussion#how-can-i-get-notified-if-a-translated-file-changes . That way you won't have to keep checking files over and over.

ltGuillaume:
Thx! I haven't seen your tutorial just yet, but actually I just get
GitHub's RSS feed for the commits of /values-nl/strings.xml :) That way, I
just have to check my RSS feeds, that's it.

ltGuillaume:
Just read it (couldn't get there easily via your link in the GitHub
notification mail somehow, but that's another good method, yeah. If you
(still) use an RSS reader, my method is easy, too.

PR #136, opened 2017-12-14
Update Korean translation

hanjoongcho:


tibbi:
Hey, thanks. I wrote a guide how to set up a file watcher which will send you an email whenever a file changes, you can check it out at https://github.com/SimpleMobileTools/General-Discussion#how-can-i-get-notified-if-a-translated-file-changes . That way you won't have to keep checking files over and over.

hanjoongcho:
I checked the link you gave me. I will set up Korean translation for receive notifications, and notifications arrive me I will translate Korean. I learned a lot through your github. Thank you.

PR #142, opened 2017-12-20
Update DE

Phoenix1747:
Translated string and changed XML entity ```&``` to ```-```.


tibbi:
no need to escape "-", just write it as it is. I already updated it. Thanks

Phoenix1747:
Oh ok, great!

PR #148, opened 2017-12-27
Updates and fixes

gregory678:
Where in the app is "unknown" string used? I don't know if I translated it correctly.


tibbi:
it was supposed to be used at contact types, like home/mobile/etc number, but there is an "Other" type which Im using. So the "unknown" isnt used anywhere yet, might not be used anytime soon :)

tibbi:
so I just started using in Contacts as the default event date (birthday, anniversary, other)

PR #155, opened 2018-01-07
Update the French translation

Poussinou:
Hi,

Everything is translated, and normally I escape all quotes ;)

However, I don't know if the French translation for the label `name="apply_to_all_apps"` will fit the button...


tibbi:
the button will stretch itself, so just hope it fits the screen :)

PR #160, opened 2018-01-09
Update strings.xml

Solatec:


tibbi:
nice, thanks for the updates!

tibbi:
in future updates just escape the apostrophes, meaning replace ' with \'

PR #234, opened 2018-02-14
confirm_otg_storage_access_text+1

10cents:
I have not translated 'OTG' though I do not believe that it is a common known acronym, at least not among Danes in general. I do not know it myself. Acronymfinder.com suggests: On-The-Go (USB 2.0 specification), On the Ground (sales) and Off the Ground (gaming) - are you talking about USB2? Maybe I could have omitted the term 'OTG' and just translate 'device' - or will it make sense for those who need to grant access to the root directory?

### DIFF
--- a/commons/src/main/res/values-da/strings.xml
+++ b/commons/src/main/res/values-da/strings.xml
@@ -264,8 +264,8 @@
 
     <!-- OTG devices -->
     <string name="otg">OTG</string>
-    <string name="confirm_otg_storage_access_text">Please choose the root folder of the OTG device on the next screen, to grant access</string>
-    <string name="wrong_root_selected_otg">Wrong folder selected, please select the root folder of your OTG device</string>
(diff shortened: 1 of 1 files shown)


tibbi:
It means On-The-Go. It can be USB 2 or 3, it is unrelated. It means usb devices connected with the device through some cable like https://images-na.ssl-images-amazon.com/images/I/61TeaPKJFDL._SL1200_.jpg

10cents:
I think the current translation is OK then. No matter if you know the meaning of 'OTG' you will understand that it is the root folder on the USB connected device. No reason not to translate it (as I see it).

tibbi:
yeah, people who need it will know what is it, no need to explain it to everyone.

10cents:
Just got another idea, why not call it 'the USB disk'? Everybody knows what that is, and besides it will always be a disk, not any other kind of devices?

tibbi:
it can be a hard drive too

10cents:
ok

PR #245, opened 2018-02-20
faq_4 title & text

10cents:
Have you updated how it works in future versions? Here is how it works on my phone:

When you add a widget to your home screen from the widget list on your device the colour configuration options appears. You will see a square at the bottom left corner, just press it to pick a new colour. You can use the slider for adjusting the alpha too.
The option to resize it appears if you drag the widget on the screen.

### DIFF
--- a/commons/src/main/res/values-da/strings.xml
+++ b/commons/src/main/res/values-da/strings.xml
@@ -305,8 +305,8 @@
         Du kan finde en guide på siden https://github.com/SimpleMobileTools/General-Discussion, eller send mig en besked på hello@simplemobiletools.com hvis du har brug for hjælp.</string>
     <string name="faq_3_title_commons">Jeg fik slettet nogle filer ved en fejl, hvordan kan jeg gendanne dem?</string>
     <string name="faq_3_text_commons">Det kan du desværre ikke. Filer er væk så snart sletningen er bekræftet, og der er ingen papirkurv.</string>
-    <string name="faq_4_title_commons">I don\'t like the widget colors, can I change them?</string>
-    <string name="faq_4_text_commons">Yep, as you drag a widget on your home screen a widget config screen appears. You will see colored squares at the bottom left corner, just press them to pick a new color. You can use the slider for adjusting the alpha too.</string>
+    <string name="faq_4_title_commons">Jeg bryder mig ikke om farverne på widget'en, kan jeg ændre dem?</string>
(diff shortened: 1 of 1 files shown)


tibbi:
What do you mean? It works exactly as you say, I havent changed it. This is just about the colors, not resizing.

10cents:
As I read the new faq I should be able to activate the colour configuration by dragging the already installed widget. I can not, I only activate resizing. But when I _add_ a widget I can adjust the colour by clicking at the square (not the squares).

10cents:
Oh, you mean as I drag a widget _onto_ your home screen, I read 'drag' as 'move'.

tibbi:
Ye I mean that initial drag, when you pull the widget from the widgets list on the home screen. On some devices clicking is enough, but sometimes you drag it. Most widgets have 2 color squares for the text and the background, only the flashlight has one.

10cents:
I better change the translation then

10cents:
BTW what du you mean with 'only the flashlight has one'?

tibbi:
the flashlight widget has 1 colored square as theres no background, other apps have 2 for changing both the text and background

10cents:
So some phones use flashlight widget, other phones use another kind of widget. And you have to develop both types?

tibbi:
I mean the Simple Flashlight app widget has 1 square, other app widgets like Simple Music Player, Calendar etc have two.

10cents:
:) Ok. I apologize for all these questions. But in the end I think it may cause better translations. Calendar widget also has 1 square :)

tibbi:
Which one? I guess theres 2 too, you just lowered the alpha so it is transparent. Try sliding the alpha bar.

10cents:
Yep, you're right, thanks!

tibbi:
ye I will redo the widget config screen someday to make it more obvious, but I dont want to waste any time on it now.

10cents:
If you are going to improve it some day I think it would be good to be able to adjust a widget already present on the screen. And maybe frame the squares so that they are both visible if some idiots as me lower the alpha so it is completely transparent.

tibbi:
ye I will add an option to change widget colors without recreating them at the "Customize colors" screen in settings. You arent an idiot, it is just bad UI :) That screen is pretty much the only one which hasnt changed since the apps were created. It surely needs some refreshment.

10cents:
Well, I admit that I am not a genuine idiot, but enough idiot to forget to use quotation marks when appropriate. :-)

PR #253, opened 2018-02-25
Update Chinese translations & delete some strings

Pzqqt:
Some strings are not necessary translations, so I deleted them.


tibbi:
you cannot just delete some strings :) If you dont want to translate them for some reason, just leave them in english. Thanks

Pzqqt:
I just think that strings like developer names or open source library names can not be translated. we should respect the original.😂

tibbi:
the additions in brackets have been added by me, just to tell people what are they about. Library names are outside the brackets.

PR #271, opened 2018-03-01
please remove this file

Gozzwip:
i cant add new path to add the strings.xml.


tibbi:
heh, I havent noticed that. I moved the file in the proper place, go to https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/res/values-az/strings.xml and press Edit to update it. Thanks

PR #314, opened 2018-03-23
Update strings.xml

mobtechpd:
lithuanian


tibbi:
I just added a new string "Tomorrow" which you missed, translate it please when you can. Thanks

PR #326, opened 2018-04-01
Snooze

10cents:
I also would like to retranslate the sentence "Postpone reminder with Snooze by" used in the calendar app, but I can not find it, neither here nor in the calendar strings.


tibbi:
it has been removed, update your apps if you still see it :) Thanks

PR #340, opened 2018-04-11
Dutch

ltGuillaume:
I removed FAQ 5, since it's basically the same question as FAQ 3. I changed the answer for FAQ 3.

### DIFF
--- a/commons/src/main/res/values-nl/strings.xml
+++ b/commons/src/main/res/values-nl/strings.xml
@@ -204,8 +204,8 @@
 
     <!-- Confirmation dialog -->
     <string name="proceed_with_deletion">Deze items verwijderen?</string>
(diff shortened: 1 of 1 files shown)


tibbi:
nono, it is not the same. One is for active recycle bin, one without. Also you should never remove any strings. Thanks

ltGuillaume:
Sorry I removed it, my bad. Either way, I'd like to make my point once more, because the current situation doesn't make sense. One states "how can I recover", the other "how can I restore", same. The answer is "no" for one, "no, unless" for the other. So one introduces the option of a trashcan, while the other doesn't. They can be answered perfectly fine with one FAQ item.

ltGuillaume:
So you're using the one FAQ for some apps and the other FAQ for other apps (which include the trashbin) then?

tibbi:
ye of course. Thats why the string is in the Commons. It is used by multiple apps.

PR #385, opened 2018-05-29
Fix typo in 'saving' string

jacklul:
Follow up from email.
Will be keeping an eye for more typos and untranslated strings when I have time


tibbi:
alright, thanks

PR #386, opened 2018-05-30
PL translation update [2018.05.30]

gregory678:
- I cannot see "Foreground color" option nowhere in apps settings (I checked in some of them, maybe it is in apps I didn't check?);
- I think FAQ #3 and #5 are duplicates.


tibbi:
ye I thought I will be using "Foreground color" but I didnt. Even though it isnt used anywhere Ill keep it there for the future. The difference between 3 and 5 is that one answer is for apps where Recycle bin will be available, one where it wont be. I could reuse the same question, but for consistency i split them too.

PR #389, opened 2018-06-04
Update strings.xml

spkprs:
Updated whole file, please can you build a apk with my lang for test?


tibbi:
you changed the whole file structure.. Please keep the same string grouping and order, also use 4 spaces instead of 2. Thanks

PR #435, opened 2018-07-06
Update strings.xml

Solatec:


tibbi:
@Solatec update multiple strings in a single pull request please, dont create a separate one for every string. Thanks

PR #438, opened 2018-07-09
Translation update

bellingeri:


tibbi:
I just accepted an earlier pull request about the italian translation so this one has conflicts, please check if there is still something that needs to be updated. Thanks

PR #454, opened 2018-07-20
hope this is ok

TacoTheDank:
just updated a few things

even if this isnt merged, i do think kapt may need to go before annotationProcessor


tibbi:
I dont want to change such things blindly, they will break the build in some cases. Ill update when the times comes. Thanks though

TacoTheDank:
yeah sorry... i do think kapt may have to go before annotationProcessor tho, cause otherwise i think the build fails, ive had that happen before

PR #476, opened 2018-08-20
Update strings.xml

MartinGSch:
Minor change in label for sorting, three dots are a good indication that a selection window will follow the selection. Added here for consistency with German label for Simple Gallery.


tibbi:
those dots are used as dialog titles too and no language has them, so I dont want them anywhere. I removed them from the gallery too.

PR #477, opened 2018-08-22
Update strings.xml

MartinGSch:
Minor language improvements, mainly punctuation

### DIFF
--- a/commons/src/main/res/values-de/strings.xml
+++ b/commons/src/main/res/values-de/strings.xml
@@ -2,14 +2,14 @@
     <string name="ok">OK</string>
     <string name="cancel">Abbrechen</string>
     <string name="save_as">Speichern unter</string>
-    <string name="file_saved">Datei erfolgreich gespeichert</string>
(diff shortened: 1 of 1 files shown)


tibbi:
there should be no punctuation and no apps have it, so lets be consistent about it. Thanks

MartinGSch:
It would be good to reconsider this... I have inserted punctuation only where it is appropriate for proper German language, i.e. at the end of complete sentences as they appear e.g. in pop up messages. That's where they are needed otherwise it just looks faulty. No dots at the end in many other items like labels, that's clear. 
I love to be consistent too, in my view these changes were just the start to make it consistently also across the other apps ;-).
If you want to go for high quality user experience and user interface I can only recommend to add dots where appropriate. At least for German where I can recommend this as a native speaker, and after cross checking official recommendations for correct and good quality language, when to leave out the dots at the end.

tibbi:
toast messages usually dont have dots, so I dont want to have them either

PR #482, opened 2018-09-12
Update strings.xml

spkprs:
Label


tibbi:
the strings are ok, no change is needed. Thanks

PR #495, opened 2018-10-02
delete_warning

10cents:
In this case I suggest not to change the string. It does not look so good if the variable will be replaced by a number, and I do not know which version of the Danish 'one' will be used if it is replaced by a word. Actually I now regret I didn't suggest the same for <item quantity="one">Moving %d item into the Recycle Bin</item>, 'Flytter en fil til papirkurven'

### DIFF
--- a/commons/src/main/res/values-da/strings.xml
+++ b/commons/src/main/res/values-da/strings.xml
@@ -232,7 +232,7 @@
     <string name="no">Nej</string>
 
     <plurals name="delete_warning">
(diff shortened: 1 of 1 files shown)


tibbi:
It is for consistency, and Android studio gives warnings that a variable should always be used. But if it complicates the translation in your case, feel free removing the variable at the Bin string too. Thanks

10cents:
Well, I will change it then, it gives a more elegant translation.

PR #509, opened 2018-10-13
Update strings.xml

Pzqqt:


tibbi:
any reason why you reverted the description of libraries at the bottom? The part between the brackets does not belong to the library name, it just explains what is which library doing. Thanks

Pzqqt:
> any reason why you reverted the description of libraries at the bottom? The part between the brackets does not belong to the library name, it just explains what is which library doing. Thanks

I think they are technical terms and should not be translated. And the translations previously submitted by others belong to machine translation, which is stupid.

tibbi:
well machine translation sucks indeed. But the terms should be translated in a casual way, nothing fancy there

PR #511, opened 2018-10-15
Slovenian translations

franga2000:
I've finally finished the Slovenian translations. Took me a while to fix the mess that Android Studio's translation editor made, but I think the formatting should be good now.

_P.S.: 
In case anyone else ever gets optimistic enough to use the editor as I did and then needs to match order and formatting with the original, [here's my (crappy) script for that](https://gitlab.com/snippets/1763809)._


tibbi:
nice, thanks :)

PR #537, opened 2018-11-02
Dutch

ltGuillaume:


ltGuillaume:
I was looking for information on what the Upgrade would entail, but I can't find anything. Would you be so kind to elaborate on your plans (or point me to where I can read about it)?

ltGuillaume:
That's completely fair, even to those who really can't spare the money. You
definitely deserve compensation for such active development on no-nonsense
functional and trustworthy apps, which has become a sparse commodity
nowadays.

PR #543, opened 2018-11-03
adding option keep both in conflicts dialog 

Doubl3MM:
Creating feature based on issue https://github.com/SimpleMobileTools/Simple-Gallery/issues/818. Adding option "keep both" to the conflict dialog that will keep both files(e.g. image.png, image(1).png)


tibbi:
yeah it seems to work fine, thanks

PR #612, opened 2019-01-19
Updated Hungarian translation

hevesij:


tibbi:
@hevesij not sure how are you finding new strings, but in case you are interested I wrote a guide about setting up emails, so in case some string gets added or updated, you get an email notification. It is at https://github.com/SimpleMobileTools/General-Discussion#how-can-i-get-notified-if-a-translated-file-changes . Thanks

PR #615, opened 2019-01-23
Dutch

ltGuillaume:
sideloaded_app? Are you introducing copy protection? :-P


tibbi:
nah, it is an issue caused by app bundling :) If a user shares his apk directly, it might not be a good version for the receiving user. Instead of running the app it will just keep crashing to him. So instead of crashing I will just show a prompt that his app version is corrupt. This wont prevent app clones from showing up, it also wont disable fdroid, yalp etc.

PR #632, opened 2019-02-06
UPD print [with notify ;-)]

spkprs:
nice, came with mail!!


tibbi:
see, it works :) I just updated another string, guess you will be notified soon. Thanks

PR #660, opened 2019-02-24
Update italian translation

ema-pe:


tibbi:
theres a few more from 384, thanks

PR #663, opened 2019-02-25
Update Japanese translation

ScratchBuild:


tibbi:
this might be wrong, you added many new string versions. I was getting warnings that the string version for "one" is the same as "other", so I removed it. Even in your new strings it is the same, so the "one" doesnt need to be there. Please create a new pull request with just the translations, no need to add the "one" versions. Thanks

PR #666, opened 2019-02-28
Update strings.xml

FTno:
Norwegian (nb) translation update


tibbi:
thanks, there are a few more from 324 and 358 :)

PR #668, opened 2019-03-03
Update Swedish translation

en2sv:


tibbi:
thanks, there are a few more from 324 :)

PR #717, opened 2019-05-07
Update german translaten

mklopp:
I am uncertain on how to escape the '&' character.
On line 662 I used &amp;
But on line 190 and 588 I did not because it is wrapped in quotes.


tibbi:
no need to add the language to the urls, just use the default links. Thanks

PR #720, opened 2019-05-15
Update strings.xml

motomoto7:
Please rename values-uk onto values-ua, uk usually is shortened for the UK, please check List of ISO 3166 country codes on wikipedia. Thanks.


tibbi:
uk is used for Ukraine by Google everywhere. Thanks

PR #724, opened 2019-05-22
Dutch

ltGuillaume:
I suppose the %? are fixed and are not dependent on whatever is in this string? So ppl can't "refactor" %Y (year) to %J (jaar) by changing it in this string?


tibbi:
yeah, you did it correctly. Thanks

PR #738, opened 2019-06-16
Update Brazilian Portuguese strings

hugomg:
Hello,

I updated the Brazilian Portuguese strings file. I believe this covers all the stuff that used to still be in english.


tibbi:
nice, thanks

PR #799, opened 2019-10-12
Dutch

ltGuillaume:


tibbi:
there are some untranslated FAQ items from line 645, you could take a look at those too. Thanks

ltGuillaume:
Just got the notification for that via RSS, will have a look.

PR #818, opened 2019-11-11
allow picking multiple files

sum12:
this should help fix https://github.com/SimpleMobileTools/Simple-Gallery/issues/1524
I might be wrong though

since this is a common lib, I am not sure how other apps will behave when multiple file selection is allowed

AND, thanks a bunch for these tools. I have been using them in excess  :)


tibbi:
This is not the view used by the gallery picker, it is the MediaAdapter in the app itself. This dialog is for example for picking the copy/move destination folder, after pressing "Other folder" at the bottom left corner. There is usually a flag inside the third party intent itself stating if selecting multiple files is allowed, or not. Thanks though

PR #831, opened 2019-12-29
Update strings.xml

AlbatorV:


tibbi:
this is the item shown at the Menu too, so I dont want to add dots there. Thanks

PR #859, opened 2020-02-13
update nl strings.xml

Joppla:
Names of days with small start letter in stead of capital according to Dutch spelling rules.

PR #860, opened 2020-02-18
Update nl string.xml

Joppla:
Made more in style of official spelling of Dutch language. (days of the week are written with small letter in Dutch, shorten form is with 2 letters). The letter "ij" in vrijdag is one symbol and cannot be splitted.

This should make the clock widget more in style of the "official" android clock widget in Dutch version.

left the "official" clock, right the SMT clock widget as it is now:
klok

If other files has to be changed to get the wished result, please let me know.

See also my comment here: [https://github.com/SimpleMobileTools/Simple-Commons/pull/859#issuecomment-585962239](url)

### DIFF
--- a/commons/src/main/res/values-nl/strings.xml
+++ b/commons/src/main/res/values-nl/strings.xml
@@ -544,13 +544,13 @@
     <string name="in_november">in november</string>
     <string name="in_december">in december</string>
 
-    <string name="monday">Maandag</string>
-    <string name="tuesday">Dinsdag</string>
-    <string name="wednesday">Woensdag</string>
-    <string name="thursday">Donderdag</string>
-    <string name="friday">Vrijdag</string>
-    <string name="saturday">Zaterdag</string>
-    <string name="sunday">Zondag</string>
+    <string name="monday">maandag</string>
+    <string name="tuesday">dinsdag</string>
+    <string name="wednesday">woensdag</string>
+    <string name="thursday">donderdag</string>
+    <string name="friday">vrijdag</string>
+    <string name="saturday">zaterdag</string>
+    <string name="sunday">zondag</string>
 
     <string name="monday_letter">M</string>
     <string name="tuesday_letter">D</string>
@@ -560,13 +560,13 @@
     <string name="saturday_letter">Z</string>
     <string name="sunday_letter">Z</string>
 
-    <string name="monday_short">Maa</string>
-    <string name="tuesday_short">Din</string>
-    <string name="wednesday_short">Woe</string>
-    <string name="thursday_short">Don</string>
-    <string name="friday_short">Vri</string>
-    <string name="saturday_short">Zat</string>
-    <string name="sunday_short">Zon</string>
+    <string name="monday_short">ma</string>
+    <string name="tuesday_short">di</string>
+    <string name="wednesday_short">wo</string>
+    <string name="thursday_short">do</string>
+    <string name="friday_short">vr</string>
+    <string name="saturday_short">za</string>
+    <string name="sunday_short">zo</string>
 
     <!-- Pro version -->
     <string name="upgrade_to_pro_long">Deze app zal niet langer worden bijgewerkt. Upgrade naar de Pro-versie om gebruik te kunnen maken van nieuwe functies en verbeteringen.</string>


tibbi:
@ltGuillaume any opinion? Is this ok?

ltGuillaume:
@tibbi Yes, these short days were added automatically, and I must have missed them. 2 characters is the standard for Dutch.

As for making it lowercase, I'm not sure @Joppla. Personally, I don't like how it looks on the clock widget, but the thing is, where else is it used? If it's ever used inside of a "sentence", such as, I don't know, "Alarm is set for _Friday_, ...", then it sure has to be lowercase in Dutch. If it's only used at the beginning of a sentence/at places like the clock widget, I would prefer a capital.

Joppla:
@ltGuillaume : Thanks for your reaction. In my opinion, as Dutch citizen (I don't know from where you are), I like more the small letters in the clock widget than the Capital. Only at the beginning of a sentence [Friday I will go to the market // = Vrijdag ga ik naar de markt] I should use the capital. In all other cases, I should use the small letter. Or ... If all letters are capital of course: MA DI WO etc. But not Ma, Di, Wo etc.

Hopefully this helps you.

ltGuillaume:
@Joppla Yeah I guess we generally agree with you.

@tibbi Are the days/short_days used in sentences as first word? If so, do they get a capital automatically? Then the changes @Joppla are great.

ltGuillaume:
@Joppla Yeah I guess I generally agree with you.

@tibbi Are the days/short_days used in sentences as first word? If so, do they get a capital automatically? Because in Simple Calendar, you have monday_alt etc. that specifically are for uses inside a sentence, so I figure the ones we're talking about here should have a capital.

Also strange: in Simple Calendar I see "Augustus 20 (do)", short day, lowercase. Where does that come from then?

Joppla:
In the Simple Calendar I see: "AUGUSTUS 20 (DO)" all in capitals. But that DO is indeed a question where it comes from.

ltGuillaume:
Screenshot_20200218-203709_Agenda
All caps? Where?

Joppla:
Here:
Calendar

tibbi:
It is mostly used in the calendar and clock. Only in such short uses like shown previously, I avoid longer sentences like "Alarm is set for Friday...". That would be too difficut to write properly in all languages.

ltGuillaume:
Then I would propose, since it's more common to use capitals in "title" contexts as opposed to sentences, to change back to a capital and to fix only the short days to 2 characters. What do you think, @Joppla?

tibbi:
those titles like "AUGUSTUS 20 (DO)" are set to uppercase in code, it is used like that in all languages. I dont want to change that, it is a UI thing.

Joppla:
> 
> 
> those titles like "AUGUSTUS 20 (DO)" are set to uppercase in code, it is used like that in all languages. I dont want to change that, it is a common UI thing.

I agree



> 
> 
> Then I would propose, since it's more common to use capitals in "title" contexts as opposed to sentences, to change back to a capital and to fix only the short days to 2 characters. What do you think, @Joppla?
I m not sure if I understand you well. But I prefer to see in the clockwidget this kind of syntax:
` vr 14 feb // vr 14 februari // VR 14 FEB // VR 14 FEBRUARI`
It disturb me less if a sentence start with a small letter (f.e. ma or maandag) in case you use the short form than to see "Ma" in the widget. Special because almost nowhere starts a sentence with the short form of a day.

ltGuillaume:
@tibbi Yeah I know, it's not the all-caps stuff I meant.

Thing is, if we lowercase the days, then this would happen:
image

...so, the months would have to become lowercase, too.

But what would happen here?
image

Would it become "februari", or is a capital forced there, just as all-caps is forced in the screenshot earlier?

@tibbi Also, like the Brits, we say "19 februari (2020)", not "Februari, 19". Is there a way of doing this? Android properly adheres to this, but it seems Simple Calendar/Clock do not.

Joppla:
@ltGuillaume 
> ...so, the months would have to become lowercase, too.

I have done before and it is accepted by @tibbi 

> Would it become "februari", or is a capital forced there, just as all-caps is forced in the screenshot earlier?

I prefer "februari" or "FEBRUARI", not "Februari"

ltGuillaume:
> I have done before and it is accepted by @tibbi

Aaah, I see. Ok, if it doesn't affect the header above the montly calendar, for instance, that wouldn't make any sense in lowercase; it's a header/title.

But it would also be better if the calendar said "19 februari (wo)" (so, lowercase), instead of "Februari 19 (wo)".

Joppla:
> But it would also be better if the calendar said "19 februari (wo)" (so, lowercase), instead of "Februari 19 (wo)".

Yes

Joppla:
And now?

tibbi:
not sure, there isnt a clear result as I see. The strings are used only in such short strings in clock and calendar, no complex sentences. The strings either begin with the week name, or it is written in brackets after the date. It feels quite weird to have them in lowercase to me, but Im not from your country.

ltGuillaume:
> not sure, there isnt a clear result as I see. The strings are used only in such short strings in clock and calendar, no complex sentences. The strings either begin with the week name, or it is written in brackets after the date. It feels quite weird to have them in lowercase to me, but Im not from your country.

I agree with you. I honestly think - like I said earlier - that the fact that these _are_ lowercase in the stock Android clock, is only because they use the translations in a sentence, too.

But what about this:
>Also, like the Brits, we say "19 februari", not "Februari 19". Is there a way of doing this in the calendar, too? Android properly adheres to this, I see _your clock does as well_, but it seems Simple Calendar does not.

tibbi:
ye adding different date/time formats is on the roadmap, but I cannot promise any date yet

ltGuillaume:
All right, that's clear. What about:

Would it become "februari" with the current translation by @Joppla (without caps), or is a capital forced there?

tibbi:
there is no capital forced here, it is usually used only at all-caps words

ltGuillaume:
> there is no capital forced here, it is usually used only at all-caps words

Errr... so that means the months become "februari" (lowercase) _in those two screenshots_ with the latest translation update by @Joppla?

PR #870, opened 2020-03-05
Update strings.xml

AlbatorV:


tibbi:
thanks, theres a few more untranslated strings starting from line 610 :)

PR #896, opened 2020-04-14
copy_to_clipboard

10cents:


tibbi:
you can change any amount of lines at once, no need to create a new PR for every update :) Thanks though

PR #898, opened 2020-04-14
changed all single quotes to double quotes and added missing translations

Honk2:
added missing translations


tibbi:
nice, thanks

PR #899, opened 2020-04-15
Update strings.xml

Solatec:


tibbi:
updated an apostrophe for consistency. The one on line 33 needs to be escaped with a backslash, the original one on 35 didnt need it. Not sure if you used 2 different ones on purpose, or it is some missclick or so.

PR #903, opened 2020-04-19
Updated Hungarian translation

hevesij:


tibbi:
you reverted a lot of already translated strings, please fix it

PR #917, opened 2020-05-07
Updating French strings

Charles-770:
I was not able to find those others corrections:
* You have to put 'Poids' instead of 'Dimension';
* And 'Dimensions' (with an 's' at the end) instead of 'Résolution' (the resolution is measured in dpi or ppi).
* Likewise, the unit of weight for images is denoted 'ko', 'Mo' or 'Go' (the 'o' in lower case) in french, instead of 'kB', 'MB', 'GB'.
 * In French, the decimal is noted by a comma instead of a dot, and a space after the value ('0,2 MP' instead of '0.2MP').

### DIFF
--- a/commons/src/main/res/values-fr/strings.xml
+++ b/commons/src/main/res/values-fr/strings.xml
@@ -32,12 +32,12 @@
     <string name="number">Numéro</string>
     <string name="no_contacts_found">Aucun contact n\'a été trouvé</string>
     <string name="request_the_required_permissions">Demander les autorisations nécessaires</string>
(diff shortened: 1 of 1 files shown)


Charles-770:
Also, in the Editor, 'Focus' should be renamed to 'Floutage'.

tibbi:
well for the gallery strings go into the gallery repository, it is at https://github.com/SimpleMobileTools/Simple-Gallery/blob/master/app/src/main/res/values-fr/strings.xml . Thanks

PR #932, opened 2020-05-25
updated ChangeDateTimeFormatDialog (for Simple-Calendar)

Honk2:
preparation for using this dialog in Simple-Calendar
* added new date format dd. MMMM yyyy (common in Germany)
* added parameter to hide 24h clock checkmark (for date format only)


tibbi:
1. what will be the use case of showing the dialog without a time?
2. do not use Today as an example. It could be confusing at dates like 7.7.2020

Honk2:
1. The date format selector should be used in the calendar app. Different location can have different selected date formats. e.g. The event list and the event edit could be different. In this case, it would be confusing to select the time format. But maybe the problem is, that the selected date format is stored in the baseconfig? In this case it would required another parameter to skip that.

2. understood, will remove that one

tibbi:
I can add the new format, no problem with that. I even asked on Facebook if someone uses a different format. Remove the optional time format setter too, lets always show it. At least we will be able to remove the checkbox in the Calendar app. I want as much consistency between apps as possible.

PR #944, opened 2020-07-05
change primary color of black theme

Honk2:
Changed the primary color from black to the dark primary color.
The primary color is used for several labels in the settings and the apps.
Therefore these texts are not visible (black on black) with the default color setting for the black and white theme.

As the theme is not black and white anymore, I suggest to rename the black and white theme to black. This is also the usual denotation for the black background theme.

Of cause changing the primary color to white will also work, but will result in a less visual structured appearance of the apps.


tibbi:
by Black & White I mean Black & White, so I dont want to change it to dark

PR #963, opened 2020-08-12
Allow custom keypad to digits converter without breaking apps

xdavidel:
Android `PhoneNumberUtils` builtin `convertKeypadLettersToDigits` converter doesn't handle well languages different then English. This patch allows sending different converters to the `highlightTextFromNumbers` extension method.


tibbi:
and which app will send a different converter? Not sure what will this change itself fix

PR #971, opened 2020-09-26
Optimize loading thumbnails

Alfavio:
[_(click to maximize image)_](https://user-images.githubusercontent.com/5478075/94331026-b9446c80-ffc9-11ea-94b6-17eda0b812b0.png)

I found a bottleneck in your app. [FilepickerItemsAdapter.kt](
https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/FilePickerDialog.kt) didn't use cached thumbnails and don't create a new cache. I fixed this in this MR. At the top is measuring the value of a metric generating new thumbnails ① and load cached thumbnails ②.


tibbi:
okay, guess it will work. Thanks

PR #978, opened 2020-10-17
Updated Hungarian translation

hevesij:


tibbi:
you reverted a lot of strings, update only english -> hungarian please

PR #984, opened 2020-11-06
Spanish translation update

FredMan95:
- The new strings that have been added were translated.
- The translation of some strings was corrected.


tibbi:
cool, thanks

PR #1010, opened 2021-01-05
Adding deletion_confirmation_with_size key

LuisAlfredo92:
Adding that key and doing some of the correspondent translations.

I'm doing this because I want to show the files size on the confirmation message (for example: 'Are you sure you want to delete 3 items? (18.34 MB)') in Simple-file-manager, and since this can be also used in other apps like Simple Gallery or Simple Music Player, I decided to do a new string here in commons instead of modifying only the string in File Manager.

But, if you want to, I can modify only File Manager string or even cancel all and do nothing


tibbi:
add it to the proper place, not at the bottom please

LuisAlfredo92:
I'm sorry, I did it with the Android Studio editor and I thought that it adjust it where I wrote the original

tibbi:
lets replace "(%s MB)" with "(%s)" please. Size formatting is handled in the code, it shows kB or GB as needed.

LuisAlfredo92:
I'm sorry, I'll be busy some time so I can't do this now. I'll close this and delete the fork until a can do it and create a new one in the future, so I won't lost updates, but thanks for the help.

PR #1014, opened 2021-01-09
Update Japanese translation

ray4423:
I think use_english_language should not be translated.
People who want to display in English may not be able to read Japanese.


tibbi:
people who dont speak japanese dont have their device language set to japanese. Thanks anyway

PR #1036, opened 2021-02-08
Dutch

ltGuillaume:
**QUESTION!**

- "during_day_at": I'm assuming that %02d:%02d is replaced with the actual hours and minutes, but:
- "during_day_at_hh_mm": Do `hh` and `mm` get replaced, or is this just a regular string?


tibbi:
hh:mm is just a regular string, it wont be replaced

PR #1109, opened 2021-07-09
Updated Tamil translation

GobinathAL:


tibbi:
nice, thanks for the updates

PR #1113, opened 2021-07-18
Use color_accent for AccentColor

erfanoabdi:
* This helps to have different default primary and accent colors in case of editing colors.xml or using RROs

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/activities/CustomizationActivity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/activities/CustomizationActivity.kt
@@ -219,7 +219,7 @@ class CustomizationActivity : BaseSimpleActivity() {
                 curTextColor = getColor(theme.textColorId)
                 curBackgroundColor = getColor(theme.backgroundColorId)
                 curPrimaryColor = getColor(theme.primaryColorId)
-                curAccentColor = getColor(R.color.color_primary)
+                curAccentColor = getColor(R.color.color_accent)
                 curAppIconColor = getColor(theme.appIconColorId)
                 curNavigationBarColor = getThemeNavigationColor(curSelectedThemeId)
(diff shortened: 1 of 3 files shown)


tibbi:
I dont understand this change, why are you editing colors.xml and what is RRO?

erfanoabdi:
RRO is [runtime overlays](https://source.android.com/devices/architecture/rros)
RROs can be installed as system apps to android system image and modify user apps res values

so we can set custom colors through overlay and have different default colors.
example: https://github.com/erfanoabdi/SimpleContactsTheme

tibbi:
I still dont see the sense of adding something like that, colors can be customized in the app settings. I dont want to support app republishing, if it is about that.

erfanoabdi:
it's not for republishing of course, it helps many many users to set custom default color scheme without editing app source code at all
it's for using other than your default colors in case anyone wanted to use Simple-Commons lib on it's code
also you are doing this yourself: https://github.com/SimpleMobileTools/Simple-Commons/blob/master/samples/src/main/res/values/colors.xml but here value of colorAccent is completely gets lost

but you are using wrong color_primary instead of color_accent in your codes anyway, which is totally wrong and may lead to broken stuffs in future


(RRO was a simple test case i wanted to just to show you)

luka177:
`I dont understand this change` Ok, and i do. this change seems OK for me, and makes some sens. So why PR was closed @tibbi ?

tibbi:
because I dont want to add code that I see no benefit of

luka177:
` I see no benefit of` and I do see a benefit, your app has half broken future, and you see no benefit of fixing it?

erfanoabdi:
i'm really sorry for you, that don't understand and don't want to understand when we try to learn you something
and you don't even want to apply the fix we provided to your broken code
THIS IS NOT HOW OPEN SOURCE WORKS

PR #1116, opened 2021-07-25
Minor fixes in Spanish strings

MarcAbonce:
This assumes that the `saving_label` is always used as a header section for saving related settings rather than some indication that something is actually saving. By the way, if this is really the case, I think many of the translations are making similar conjugation errors with these settings headers.


tibbi:
ye, thats the way it is used, not in a sentence. Thanks

PR #1123, opened 2021-08-07
Fixed nonclickable links in FAQ

Aga-C:
Hi,

I've found out that in the FAQ section all links and email addresses weren't clickable, and it also wasn't possible to copy them to clipboard. It was a simple fix to do, so I've added the possibility to click on them.
The change was tested locally on Simple Thank You app, and it was working correctly.


tibbi:
looks good, thanks

PR #1133, opened 2021-08-20
Allow renaming files by just changing a letter casing

KryptKode:
**Notes**
- Allow renaming files by just changing a letter casing
- when renaming, a temporary file/directory is created and the old file/directory is renamed to the temporary file/directory, thereafter, the temporary file/directory is renamed to the new file/directory.

**Screenshots**
Screenshot

> There's an issue when renaming on SD cards, using the [DocumentsContract.renameDocument](https://github.com/KryptKode/Simple-Commons/blob/ref/allow-case-name-change/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt#L673) method, when just the case of the file name changes, it appends an integer to the name. For example, `hello` becomes `Hello (1)` when trying to rename to `Hello`  No workaround has been identified for now.  See the screenshot below. 

Screenshot

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/RenameItemDialog.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/RenameItemDialog.kt
@@ -63,12 +63,12 @@ class RenameItemDialog(val activity: BaseSimpleActivity, val path: String, val c
 
                         val newPath = "${path.getParentPath()}/$newName"
 
-                        if(path == newPath){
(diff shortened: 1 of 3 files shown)


tibbi:
do not disable the Positive button for consistency please, we are not doing it in any app

KryptKode:
> do not disable the Positive button for consistency please, we are not doing it in any app and it is unrelated to the task

Done

tibbi:
should be good, just dont forget formatting the code style. Thanks

PR #1134, opened 2021-08-24
Add support for biometric ID in SecurityDialog when targeting API 30+

qwertyfinger:
Modifies SecurityDialog to use androidx biometric library if the `targetSdkVersion` is 30+.

### DIFF
--- a/commons/build.gradle
+++ b/commons/build.gradle
@@ -35,6 +35,7 @@ dependencies {
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
     implementation 'joda-time:joda-time:2.10.3'
     implementation "androidx.exifinterface:exifinterface:1.3.2"
+    implementation "androidx.biometric:biometric-ktx:1.2.0-alpha03"
 
     api 'androidx.appcompat:appcompat:1.2.0'
     api 'com.github.ajalt.reprint:core:3.3.2@aar'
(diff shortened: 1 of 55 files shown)


qwertyfinger:
You can check out the update Notes version that uses this commons version here: https://github.com/qwertyfinger/Simple-Notes/tree/feature/biometric-lock-api30

tibbi:
1. the top dialog label still shows "Fingerprint", lets replace it with "Biometry" or something like that
2. when a note is locked and you click "Show content", or "Unlock note" lets skip our dialog and show the system one right away.
Otherwise it looks great, looks like it wont need an extra permission either.

qwertyfinger:
> 
> 
>     1. the top dialog label still shows "Fingerprint", lets replace it with "Biometry" or something like that
> 
>     2. when a note is locked and you click "Show content", or "Unlock note" lets skip our dialog and show the system one right away.
>        Otherwise it looks great, looks like it wont need an extra permission either.

Thanks, will make those adjustments 👍

qwertyfinger:
@tibbi Updated this and the example Notes branch as well: https://github.com/qwertyfinger/Simple-Notes/tree/feature/biometric-lock-api30

tibbi:
seems to work just fine, Ill test it in other apps too. Thanks!

qwertyfinger:
> 
> 
> seems to work just fine, Ill test it in other apps too. Thanks!

I tested the File Manager – worked fine. We could even show the new biometric experience for other apps too if we wanted to.

tibbi:
let's use the previous "Fingerprint" tab title on target SDK 29, it always shows Biometrics now

qwertyfinger:
> 
> 
> let's use the previous "Fingerprint" tab title on target SDK 29, it always shows Biometrics now

Ok, I will update it later in a separate PR then

PR #1142, opened 2021-09-08
Add MyRecyclerViewListAdapter that extends from ListAdapter

KryptKode:
**Notes**
- Add `MyRecyclerViewListAdapter` that extends from [`ListAdapter`](https://developer.android.com/reference/androidx/recyclerview/widget/ListAdapter)
- would be useful for some apps, so calling `notifyDataSetChanged`  multiple times does not affect performance
- it is used currently in the multiple timers implementation for the Simple Clock app [here](https://github.com/SimpleMobileTools/Simple-Clock/pull/285)


tibbi:
where will this be used?

KryptKode:
> where will this be used?

It is used in the multiple timers for Clock

PR #1145, opened 2021-09-20
update for SMS messenger  backup/restore

KryptKode:
**Notes**
- make the projection argument nullable in `Context.queryCursor `extension method
- add null equivalent methods for getLong and getString for `Cursor`
- Add `core-ktx` as api dependency and 
- update gson to `2.8.8`


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
```diff
@@ -338,7 +338,7 @@ fun Context.getMediaContent(path: String, uri: Uri): Uri? {
 
 fun Context.queryCursor(
     uri: Uri,
-    projection: Array<String>,
+    projection: Array<String>? = null,
```
tibbi, line 341: is null used anywhere, or only for debugging? We should always specify what fields are we interested in, not query everything

PR #1153, opened 2021-09-24
Add method to determine if a contact has a particular number

KryptKode:
- Add `SimpleContact.doesHavePhoneNumber` method to check if a contact has a particular number
- it performs an equality check, so the numbers match exactly
- the method is used on the `Simple SMS Messenger` app to find contacts that match a particular sender's number
- should address this [issue](https://github.com/SimpleMobileTools/Simple-SMS-Messenger/issues/128) in [this](https://github.com/SimpleMobileTools/Simple-SMS-Messenger/pull/211) pull request


tibbi:
cool, thanks

PR #1192, opened 2021-11-10
Implement handling RecoverableSecurityException 

KryptKode:
**Notes**
- Implement handling `RecoverableSecurityException`  needed for updating files in `MediaStore` 
- fix `getTempFile` method, use `folderName` parameter
- needed to solve[ this issue](https://github.com/SimpleMobileTools/Simple-Music-Player/issues/298)
- this functionality is needed in the [Simple Music Player](https://github.com/SimpleMobileTools/Simple-Music-Player/pull/353) app.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
@@ -1194,7 +1194,7 @@ fun AppCompatActivity.showSideloadingDialog() {
 }
 
 fun BaseSimpleActivity.getTempFile(folderName: String, fileName: String): File? {
-    val folder = File(cacheDir, fileName)
+    val folder = File(cacheDir, folderName)
     if (!folder.exists()) {
         if (!folder.mkdir()) {
             toast(R.string.unknown_error_occurred)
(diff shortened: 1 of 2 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
```diff
@@ -1194,7 +1194,7 @@ fun AppCompatActivity.showSideloadingDialog() {
 }
 
 fun BaseSimpleActivity.getTempFile(folderName: String, fileName: String): File? {
-    val folder = File(cacheDir, fileName)
+    val folder = File(cacheDir, folderName)
```
tibbi, line 1197: why is this needed? Wont it break something?

KryptKode: It would not break anything because of the way it's used. It was needed to [import blocked numbers](https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/kotlin/com/simplemobiletools/commons/activities/ManageBlockedNumbersActivity.kt#L119) in the dialer app. Now, it would be used in the [Music app](https://github.com/KryptKode/Simple-Music-Player/blob/feat/edit-metadata/app/src/main/kotlin/com/simplemobiletools/musicplayer/helpers/TagHelper.kt#L41) to store a copy of music files temporarily when writing tags to them.

KryptKode: There was an error in the method. It was using the `fileName` instead of the `folderName` so it's corrected here

tibbi: okay, lets hope

PR #1202, opened 2021-11-14
handle scoped storage changes

KryptKode:
**Notes**
- handle access to  `Android/[data|obb]`  with the Storage Access Framework due to scoped storage restrictions on devices running on Android 11 and above (>= Android 11)
- access to SD card and OTG generally on >= Android 11 is restricted so using SAF would not work (since we cannot select the root directory) so this PR adds a condition (`!isRPlus()`) to ensure compatibility. 
- part of the pull requests to address [this](https://github.com/SimpleMobileTools/Simple-File-Manager/issues/490)

**TODOs**
- update  [WritePermissionDialog](https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/WritePermissionDialog.kt) images to account for selecting `Android/[data|obb]`
- update error messages, [here](https://github.com/KryptKode/Simple-Commons/blob/feat/android-data-storage/commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt#L240) and [here](https://github.com/KryptKode/Simple-Commons/blob/feat/android-data-storage/commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt#L240) when one selects the wrong directory for `Android/[data|obb]`


tibbi:
damn, didnt expect it to be this big change :/ Ill have to do a lot of testing, then release it really carefully.

tibbi:
updated some code styles, I'm afraid of touching the rest :D

tibbi:
anyway, not much to test within Simple Commons so Ill accept it and test in some other apps that really work with files

PR #1227, opened 2021-11-29
Fixed contact padding

Aga-C:
Hi,

I've found in Simple Contacts, that when you hide contact thumbnail and show number, there is no margin between the side of the screen and text. I did the same paddings as there are for contacts without displaying phone number.

**Before:**

Screenshot_2

**After:**

Screenshot_1


tibbi:
how about if you enable thumbnails? Doesnt this add extra padding?

Aga-C:
I haven't checked, there was a problem. But it was easy to fix. Now it looks like this:

image

image

tibbi:
ye looks good, thanks

PR #1230, opened 2021-12-04
Add ChecksSdkIntAtLeast to Sdk version checks

ebraminio:
The linter is able to infer this inside a project but
in order to make it work more reliably cross project
it is better to make it explicit.

Found this is needed on https://github.com/SimpleMobileTools/Simple-Clock/blob/c812483f132eb3a8d67f1c481e7e3399c77b7a7e/app/src/main/kotlin/com/simplemobiletools/clock/services/TimerService.kt#L94 where `isOreoPlus` is stated by liner can't understand it thus `@TargetApi(Build.VERSION_CODES.O)` is used to silence the complain.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/helpers/Constants.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/helpers/Constants.kt
@@ -6,6 +6,7 @@ import android.graphics.drawable.Drawable
 import android.os.Build
 import android.os.Looper
 import android.util.Log
+import androidx.annotation.ChecksSdkIntAtLeast
 import com.simplemobiletools.commons.R
 import com.simplemobiletools.commons.overloads.times
 import java.util.HashMap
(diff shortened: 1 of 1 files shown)


tibbi:
this isnt a cross project, it is supposed to be used in our apps only

ebraminio:
> this isnt a cross project, it is supposed to be used in our apps only

Yeah, I wanted to use on your other project not somewhere else. Perhaps I didn't explain it enough,

In [here](https://github.com/SimpleMobileTools/Simple-Clock/blob/c812483f132eb3a8d67f1c481e7e3399c77b7a7e/app/src/main/kotlin/com/simplemobiletools/clock/services/TimerService.kt#L94) you have `if (isOreoPlus()) {` but if you had `Build.VERSION.SDK_INT >= Build.VERSION_CODES.O` on the project itself linter could recognize that part of the code is already sdk version limited so instead `@TargetApi(Build.VERSION_CODES.O)` is put there.

tibbi:
I see, will test it later. Thanks

tibbi:
ok, seems to be good, thanks

PR #1232, opened 2021-12-06
Translate to Spanish the words in English

AguanteDebian:
I think I translated all the fileto Spanish. If it's any word left without translating or any error in the programming grammatical, let my know.


tibbi:
looks good, thanks :)

PR #1240, opened 2021-12-15
Added date to settings backup

Aga-C:
Closes SimpleMobileTools/Simple-Gallery#2078

Screenshot_1


AlbatorV:
Thanks
Maybe format could be "gallery-settings_20211215_123830.txt"

Aga-C:
@tibbi I noticed now that this date won't change after someone saves the file because the name is remembered. Can I remove remembering previous name, or maybe should it be done in another way?

tibbi:
lets just always generate a new name without remembering, will see how it goes

PR #1242, opened 2021-12-19
Update strings.xml

AlbatorV:
Updated... even if I don't understand why app's names should be translated.


tibbi:
thanks. Well, because they dont make much sense in other languages.. It isnt some abstract name like Google

PR #1245, opened 2021-12-20
Feat/bottom cab

KryptKode:
**Notes**
- Implement bottom contextual action mode
- Sample usage can be seen on the Simple App Launcher in [this PR](https://github.com/SimpleMobileTools/Simple-App-Launcher/pull/129/)

**Migration for adapters**
- Remove the `prepareActionMode(menu: Menu)` and `onActionModeCreated` methods and implement the new `onBottomActionMenuCreated(view: BottomActionMenuView)` which serves the same purpose as `prepareActionMode`; items properties can be changed here. 

**PS**
Could not find a way to adapt the background color to the theme background color programmatically

https://user-images.githubusercontent.com/25648077/146713737-900e8de3-586c-4510-b613-2e968e0878d8.mp4


tibbi:
nice, will test it soon. Can you rename it to "BottomActionMenu" or so though please? ContextView doesnt really explain what is it about.

commons/src/main/kotlin/com/simplemobiletools/commons/views/bottomactionmenu/BottomActionMenuPopup.kt
```diff
@@ -0,0 +1,74 @@
...
+import android.view.accessibility.AccessibilityEvent
+import android.widget.PopupWindow
+import androidx.annotation.MenuRes
+import androidx.core.widget.PopupWindowCompat
+import com.google.android.material.floatingactionbutton.FloatingActionButton
+import com.simplemobiletools.commons.activities.BaseSimpleActivity
+
+
+class BottomActionMenuPopup(private val activity: BaseSimpleActivity, items: List<BottomActionMenuItem>) {
+    private val contextView = BottomActionMenuView(activity)
```
tibbi, line 15: rename these too please

commons/src/main/res/values/attrs.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <declare-styleable name="ContextViewItem">
```
tibbi, line 3: these too

samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/StringsAdapter.kt
```diff
@@ -41,7 +41,7 @@ class StringsAdapter(
 
     override fun getActionMenuId() = R.menu.cab_delete_only
 
-    override fun prepareActionMode(menu: Menu) {}
+    override fun onContextViewCreated(view: ContextView) {}
```
tibbi, line 44: still says ContextView here

commons/src/main/res/values/dimens.xml
```diff
@@ -31,6 +34,9 @@
     <dimen name="rounded_corner_radius_big">10dp</dimen>
     <dimen name="bottom_sheet_corner_radius">24dp</dimen>
     <dimen name="breadcrumbs_layout_height">48dp</dimen>
+    <dimen name="bottom_cab_layout_height">48dp</dimen>
+    <dimen name="bottom_cab_layout_max_width">300dp</dimen>
+    <dimen name="bottom_cab_layout_item_size">48dp</dimen>
```
tibbi, line 39: some of these newly added dimensions arent used even in App Launcher, I dont think they are needed, right?

commons/src/main/res/values/dimens.xml
```diff
@@ -20,7 +20,10 @@
     <dimen name="colorpicker_color_width">70dp</dimen>
 
     <dimen name="normal_icon_size">48dp</dimen>
+    <dimen name="fastscroll_width">8dp</dimen>
+    <dimen name="fastscroll_height">40dp</dimen>
     <dimen name="fingerprint_icon_size">72dp</dimen>
+    <dimen name="fastscroll_bubble_offset">30dp</dimen>
```
tibbi, line 26: old fastscroller values arent used anymore either

commons/src/main/res/values/dimens.xml
```diff
@@ -48,4 +54,10 @@
     <dimen name="big_text_size">18sp</dimen>
     <dimen name="actionbar_text_size">20sp</dimen>
     <dimen name="extra_big_text_size">22sp</dimen>
+
+    <dimen name="cab_popup_menu_vertical_padding">8dp</dimen>
+    <dimen name="cab_popup_menu_max_width">336dp</dimen>
+    <dimen name="cab_popup_menu_min_width">168dp</dimen>
+    <dimen name="cab_popup_menu_width_unit">56dp</dimen>
+    <dimen name="cab_item_min_width">85dp</dimen>
```
tibbi, line 62: these numbers are quite arbitrary, do we really need them? Menu should occupy whole width

tibbi:
it looks nice, but It behaves a bit weird. For example if I use the blocked numbers activity for testing and the menu from https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/res/menu/cab_blocked_numbers.xml , then I set the cab_delete to Never be shown, it is actually the one being shown as an icon and the copy_number is hidden at the 3 dots menu. The behaviour seems to be flipped there. Lets try fixing this one, let the hardcoded dimension values there, guess you had a reason for adding them.

tibbi:
Another thing is that the new menu hides the bottom item in the list, can we maybe do some trick there so that the menu actually pushes the list up so that the last item is visible too? Like at https://play.google.com/store/apps/details?id=com.alphainventor.filemanager . I dont want anything too hacky, maybe we can use the recyclerview reference that we are already passing to adapters like at https://github.com/SimpleMobileTools/Simple-App-Launcher/blob/1e7e8b80ff9f2ad64119f902530e4d15194ae5dc/app/src/main/kotlin/com/simplemobiletools/applauncher/adapters/LaunchersAdapter.kt#L34 ? And maybe add some bottom margin or padding to it, that shouldnt break too many things.

commons/src/main/kotlin/com/simplemobiletools/commons/views/bottomactionmenu/BottomActionMenuItemPopup.kt
```diff
@@ -0,0 +1,247 @@
...
+                return maxHeight
+            }
+        }
+
+        // At this point, we went through the range of children, and they each
+        // completely fit, so return the returnedHeight
+        return returnedHeight
+    }
+
+
```
tibbi, line 215: remove such double empty lines please and dont forget formatting the files with Ctrl + Shift + L

tibbi:
okay, this seems to be acceptable now, I will do more testing in a real app. Thanks!

PR #1248, opened 2021-12-30
Feat/bottom cab changes

KryptKode:
**Notes**
- only update the bottom action view if its item changed to fix some UI glitches
- add an  `underlayView` parameter to the `BottomActionMenuPopup#show` method which will adjust the view's bottom margin.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/views/bottomactionmenu/BottomActionMenuPopup.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/views/bottomactionmenu/BottomActionMenuPopup.kt
@@ -1,11 +1,14 @@
 package com.simplemobiletools.commons.views.bottomactionmenu
 
 import android.view.Gravity
+import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
 import android.view.accessibility.AccessibilityEvent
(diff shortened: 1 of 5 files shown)


tibbi:
the bar seems to be higher with this update, I liked the previous one better

commons/src/main/kotlin/com/simplemobiletools/commons/views/bottomactionmenu/BottomActionMenuPopup.kt
```diff
@@ -68,6 +81,24 @@ class BottomActionMenuPopup(private val activity: BaseSimpleActivity, items: Lis
...
+                        bottomMargin - bottomActionMenuView.height
+                    }
+
+                    if (newMargin >= 0) {
+                        bottomMargin = newMargin
+                    }
+                }
+            }
+        }
+    }
```
tibbi, line 100: Is this whole height thing added for being able to see the last item in the listview? Because it makes the whole bottom bar higher by default now, I liked the previous height better.

KryptKode: Will revert it.

PR #1250, opened 2021-12-31
change Bottom Action Menu icon colors to white

KryptKode:
**Notes**
change Bottom Action Menu icon colors to white


tibbi:
cmon man, just use Color.WHITE, do not complicate things unnecessarily

tibbi:
yeah, this fix itself is good, but I found 3 more issues at pressing the 3 dots. 
1. If the menu label takes more than 1 line, that whole menu is not centered, it is scrolling and it feels weird . Make the menu high enough to not scroll, do not hardcode its height.
2. there seems to be some small margin or padding around the menu item and when I click it, the ripple effect doesnt go all the way to the side.
3. lets not show icons there at the 3 dots menu, it will look weird if some labels will have an icon and some wont.
Screenshot_20220103-094413

PR #1286, opened 2022-02-05
Revert from using the bottom action menu

KryptKode:
**Notes**
- This revert from using the previously added bottom action menu. 
- Apps will now use the top contextual action menu
- This is done because the bottom action menu implementation is not complete yet and apps need to be updated


tibbi:
StringsAdapter wasnt updated, always compile the code before commiting, to make sure it works. Thanks

PR #1306, opened 2022-02-15
Feat/deleting with SAF on SDK 30+

KryptKode:
### Deleting on Apps on SDK 30+ using Storage Access Framework (SAF)
- refactor delete for SDK 30+ to use the SAF for files in all directories except the directories specified [here](https://developer.android.com/training/data-storage/shared/documents-files#document-tree-access-restrictions)
- this aims to prevent prompting the user every time they want to delete a file. 
- with this change, the user is only prompted the first time to select the first parent folder where the file is present. 
- the tree URI of the first parent folder is used subsequent times if the permission is still persisted. 
- TODO: update the image in the `WritePermissionDialog`

### Fix prompt for multiples files when using `MediaStore.createDeleteRequest`
- fix deleting multiple files prompting for each file one by one when using the SDK 30+'s `MediaStore.createDeleteRequest`.
-  add a new parameter `isDeletingMultipleFiles: Boolean` to the `deleteFileBg` method to perform the check. 

### New string toast message
- add a new string `wrong_folder_selected`
- use the string for when a user selects the wrong SAF tree for delete
- use the string as a message for when the user selects the wrong tree when they are supposed to select `Android/[data|obb]` tree.


commons/src/main/res/values/strings.xml
```diff
@@ -228,6 +228,7 @@
     <string name="sd_card_usb_same">SD card and USB device paths cannot be the same</string>
     <string name="app_on_sd_card">You seem to have the app installed on an SD card, that makes the app widgets unavailable. You won\'t even see them on the list of available widgets.
         It is a system limitation, so if you want to use the widgets, you have to move the app back on the internal storage.</string>
+    <string name="wrong_folder_selected">Wrong folder selected, please select the path %s</string>
```
tibbi, line 231: please add apostrophes around %s to make it \'%s\' and remove "the", just "please select \'%s\'

commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context-storage.kt
```diff
@@ -155,12 +155,24 @@ fun Context.isPathOnInternalStorage(path: String) = internalStoragePath.isNotEmp
 
 private const val ANDROID_DATA_DIR = "/Android/data/"
 private const val ANDROID_OBB_DIR = "/Android/obb/"
+private const val DOWNLOAD_DIR = "Download"
```
tibbi, line 158: this needs more research, at renaming it is both Download and Documents + their subfolders behaving differently, not just Download itself

tibbi:
added just a couple styling notes for now, will have to test real functionality in an app like Gallery or File Manager

commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context-storage.kt
KryptKode: On investigating, we can select tree URI of all directories except
-  the root of Internal storage,
-  the root of SD card and, 
- the Download folder of either the Internal Storage or SD card

The information is [here](https://developer.android.com/training/data-storage/shared/documents-files#document-tree-access-restrictions)

PR #1323, opened 2022-02-24
Feat/deleting with SAF sdk30+

KryptKode:
### Notes
- some cleanups to the process of handling delete for SDK 30+

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context-storage.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context-storage.kt
@@ -164,6 +164,10 @@ fun Context.getSAFOnlyDirs(): List<String> {
 }
 
 fun Context.isAccessibleWithSAFSdk30(path: String): Boolean {
+    if(path.startsWith(filesDir.absolutePath)){
+        return false
+    }
+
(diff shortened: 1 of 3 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context-storage.kt
```diff
@@ -164,6 +164,10 @@ fun Context.getSAFOnlyDirs(): List<String> {
 }
 
 fun Context.isAccessibleWithSAFSdk30(path: String): Boolean {
+    if(path.startsWith(filesDir.absolutePath)){
```
KryptKode, line 167: This check helps to exclude files already in the recycle bin. 

 I was thinking of bringing the [`recycleBinPath` defined in the Gallery module](https://github.com/SimpleMobileTools/Simple-Gallery/blob/master/app/src/main/kotlin/com/simplemobiletools/gallery/pro/extensions/Context.kt#L73) to the commons module. Do you think that is a better approach?

tibbi: if you mean the path only and not some huge amount of related code, Im fine with it

PR #1335, opened 2022-03-13
SDK 30 changes

KryptKode:
**Notes**
- rename dialog to select SAF folder to `handleSAFDialogSdk30`
- use SAF to copy/move to/from folders accessible with SAF
- this involves creating new files and folders
- change error message while creating a new folder
- change the constant name for starting document tree intent to `OPEN_DOCUMENT_TREE_FOR_SDK_30`
- fix the bug that caused the function not to run after selecting SAF folder


tibbi:
will accept it here and test in a real app

PR #1340, opened 2022-03-15
sdk 30 changes - Allow creation of folder in root storage

KryptKode:
## Notes
- handle creating a new folder in the root of internal or SD card storage
- use `ACTION_CREATE_DOCUMENT` intent with SAF to give the user opportunity to save the desired folder. 

https://user-images.githubusercontent.com/25648077/158484258-95519c78-6248-409f-8f58-9cbf5db0477e.mp4

### NB: Would need to update the image in the `WritePermissionDialog` for this usecase

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt
@@ -223,8 +223,27 @@ abstract class BaseSimpleActivity : AppCompatActivity() {
         }
 
         val sdOtgPattern = Pattern.compile(SD_OTG_SHORT)
+        if (requestCode == CREATE_DOCUMENT_SDK_30) {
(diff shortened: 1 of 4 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt
```diff
@@ -430,7 +449,20 @@ abstract class BaseSimpleActivity : AppCompatActivity() {
...
+            callback(true)
+            false
+        }
+    }
+
+    fun handleSAFCreateDocumentDialogSdk30(path: String, callback: (success: Boolean) -> Unit): Boolean {
+        return if (!packageName.startsWith("com.simplemobiletools")) {
+            callback(true)
+            false
+        } else if (isShowingSAFCreateDocumentDialogSdk30(path)) {
```
tibbi, line 465: whats this? Such function doesnt exist, it cannot be compiled

PR #1343, opened 2022-03-19
sdk 30 changes

KryptKode:
## Changes
- remove `handleSAFDialogSdk30` for source paths since we have READ permission
- display the desired path in `WritePermissionDialog` for selecting tree URI on SDK 30+

https://user-images.githubusercontent.com/25648077/159136292-d333126e-dde2-486d-8286-a0395e4a8ec5.mp4

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/WritePermissionDialog.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/WritePermissionDialog.kt
@@ -11,39 +11,39 @@ import kotlinx.android.synthetic.main.dialog_write_permission.view.*
 import kotlinx.android.synthetic.main.dialog_write_permission_otg.view.*
 
 class WritePermissionDialog(activity: Activity, val mode: Mode, val callback: () -> Unit) {
-    enum class Mode {
-        OTG,
-        SD_CARD,
-        OPEN_DOCUMENT_TREE_SDK_30,
-        CREATE_DOCUMENT_SDK_30,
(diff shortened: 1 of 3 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/WritePermissionDialog.kt
```diff
@@ -11,39 +11,39 @@ import kotlinx.android.synthetic.main.dialog_write_permission.view.*
...
-    enum class Mode {
-        OTG,
-        SD_CARD,
-        OPEN_DOCUMENT_TREE_SDK_30,
-        CREATE_DOCUMENT_SDK_30,
+    sealed class Mode {
+        object Otg : Mode()
+        object SdCard : Mode()
+        data class OpenDocumentTreeSDK30(val path: String) : Mode()
+        object CreateDocumentSDK30 : Mode()
```
tibbi, line 18: what are these sealed, object etc? Looks weird

tibbi: is it supposed to be some enums?

KryptKode: Only the OpenDocumentTreeSDK30 case needs a path. 
I didn't want to pass something like a null or an empty string for the other cases.

tibbi:
alright, lets see then :)

PR #1349, opened 2022-03-22
fix: renaming with just case changing

KryptKode:
## Notes
 - The MediaStore DATA column is case insensitive so renaming files with just a change in the casing of the file name still points to the same DATA column from the MediaStore’s perspective.

- So we do not need to delete the old path from MediaStore if it’s just the casing changed in the file name.


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
```diff
@@ -969,7 +969,9 @@ fun BaseSimpleActivity.renameFile(
                     runOnUiThread {
                         callback?.invoke(true, Android30RenameFormat.NONE)
                     }
-                    deleteFromMediaStore(oldPath)
+                    if(!oldPath.equals(newPath, true)){
```
tibbi, line 972: spaces please, always format the code

PR #1389, opened 2022-04-29
add floating contextual action menu at the blocked numbers

pavelpoley:


samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/MainActivity.kt
```diff
@@ -21,7 +22,8 @@ class MainActivity : BaseSimpleActivity() {
         setContentView(R.layout.activity_main)
         appLaunched(BuildConfig.APPLICATION_ID)
 
-        //startCustomizationActivity()
+        startActivity(Intent(this,ManageBlockedNumbersActivity::class.java))
+        startCustomizationActivity()
```
tibbi, line 26: do not add these changes into the PR, neither the manifest one

commons/src/main/res/layout/item_manage_blocked_number.xml
```diff
@@ -13,7 +14,21 @@
...
         android:layout_centerVertical="true"
-        android:layout_marginTop="@dimen/medium_margin" />
+        android:layout_marginTop="@dimen/medium_margin"
+        android:layout_marginEnd="16dp"
+        android:layout_toStartOf="@+id/manage_blocked_number_overflow_menu" />
+
+    <ImageView
+        android:id="@+id/manage_blocked_number_overflow_menu"
+        android:layout_width="32dp"
+        android:layout_height="32dp"
```
tibbi, line 26: use defined dimensions for this so we can easily reuse them in other cases too

commons/src/main/kotlin/com/simplemobiletools/commons/adapters/ManageBlockedNumbersAdapter.kt
```diff
@@ -69,9 +70,53 @@ class ManageBlockedNumbersAdapter(
...
+            menu.add(view.context.getString(R.string.delete))
+                .setOnMenuItemClickListener {
+                    executeItemMenuOperation(blockedNumberId) {
+                        deleteSelection()
+                    }
+                    return@setOnMenuItemClickListener true
+                }
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
```
tibbi, line 107: there are helper functions like isNougatPlus etc in Constants.kt for these

tibbi:
1. the button is way too small, difficult to click. Make the whole area above, below and to the right of the button clickable
2. add some right margin to the menu itself that appears on click, it shouldnt touch the right side of the screen
3. fix theming, the menu should use the same colors as the overflow menu that shows up at pressing the 3 dots at the top right corner in all themes

pavelpoley:
Pushed corrections
photo1651343788

tibbi:
looks better now, but it still needs some theming fixes on Android 12+ with System default (Material You) theme

tibbi:
getting there, but you removed the rounded corners in all themes now

tibbi:
right, looking good now, just dont forget formatting .xml files with ctrl + shift + L too. One more thing, please add as many things as possible from manage_blocked_number_overflow_menu and popup_anchor into a style, so that it can be added easily on other places too. I mean something like PinNumberStyle.Base in styles.xml, that will contain the paddings, centering etc. Call it something like OverflowMenuIcon and OverflowMenuAnchor.

tibbi:
also rename manage_blocked_number_overflow_menu and popup_anchor to something like overflow_menu_icon and overflow_menu_anchor, so that it is easier to copypaste to other places :x

pavelpoley:
> right, looking good now, just dont forget formatting .xml files with ctrl + shift + L too. One more thing, please add as many things as possible from manage_blocked_number_overflow_menu and popup_anchor into a style, so that it can be added easily on other places too. I mean something like PinNumberStyle.Base in styles.xml, that will contain the paddings, centering etc. Call it something like OverflowMenuIcon and OverflowMenuAnchor.

I think that we cannot extract padding and alignment, because it will fit only inside the `RelativeLayout` view group. We can extract `srcCompat` and `background`.

tibbi:
it is a padding of the view itself, doesnt matter what is the parent view

tibbi:
alright, looks good now, thanks

PR #1391, opened 2022-04-30
Feat/manage media

KryptKode:
## Notes
- add `BaseSimpleActivity.checkManageMediaOrHandleSAFDialogSdk30` used in delete/rename operations
    - this checks for whether the app is should use Storage Access Framework (`SAF`) or `MediaStore` request
    - on SDK 31, if the app can manage media, the `MediaStore` request method would be used
    - on SDK 31, if the app cannot manage media, then the `SAF` method is used
    - on SDK 30, the SAF is used as the `MANAGE_MEDIA` permission was added in SDK 31
- revert changes in `RenamePatternTab` and `RenameSimpleTab` and use the `checkManageMediaOrHandleSAFDialogSdk30` method

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
@@ -720,9 +720,9 @@ fun BaseSimpleActivity.deleteFilesBg(files: List<FileDirItem>, allowDeleteFolder
             return@handleSAFDialog
         }
 
-        handleSAFDialogSdk30(firstFile.path) {
+        checkManageMediaOrHandleSAFDialogSdk30(firstFile.path) {
             if (!it) {
-                return@handleSAFDialogSdk30
+                return@checkManageMediaOrHandleSAFDialogSdk30
             }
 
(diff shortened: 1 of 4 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
```diff
@@ -975,7 +983,7 @@ private fun BaseSimpleActivity.renameCasually(
     val tempFile = try {
         createTempFile(oldFile) ?: return
     } catch (exception: Exception) {
-        if (isRPlus() && exception is FileSystemException) {
+        if (isRPlus() && exception is java.nio.file.FileSystemException) {
```
KryptKode, line 986: This is what was causing the renaming issue. 
We need to use the full path here (or import alias). This is because we have the star import for `java.util.*`, so Kotlin auto imports its `kotlin.io.FileSystemException` here and it conflicts with the `java.nio.file.FileSystemException` that is expected

tibbi: ah okay

PR #1404, opened 2022-05-13
SimpleContact sorting

pavelpoley:
Added sort by date created + ASC/DESC.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/models/SimpleContact.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/models/SimpleContact.kt
@@ -16,7 +16,9 @@ data class SimpleContact(
     }
 
     override fun compareTo(other: SimpleContact): Int {
-        if (sorting == -1) return compareByFullName(other)
+        if (sorting == -1) {
+            return compareByFullName(other)
+        }
 
         var result = when {
             sorting and SORT_BY_FULL_NAME != 0 -> compareByFullName(other)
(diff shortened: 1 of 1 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/models/SimpleContact.kt
```diff
@@ -3,12 +3,38 @@ package com.simplemobiletools.commons.models
...
+    companion object {
+        var sorting = -1
+    }
+
     override fun compareTo(other: SimpleContact): Int {
+        if (sorting == -1) return compareByFullName(other)
+
+        var result = when {
+            sorting and SORT_BY_FULL_NAME != 0 -> {
+                compareByFullName(other)
```
tibbi, line 23: compact lines like this to 1 line please, like "sorting and SORT_BY_FULL_NAME != 0 -> compareByFullName(other)"

commons/src/main/kotlin/com/simplemobiletools/commons/models/SimpleContact.kt
```diff
@@ -27,6 +53,12 @@ data class SimpleContact(
         }
     }
 
+    private fun compareUsingIds(other: SimpleContact): Int {
```
tibbi, line 56: this function looks really useles, just add the comparison into the code

tibbi:
moved returning to a separate line, that shortening was only for when cases

PR #1406, opened 2022-05-14
fix unable to delete files in recycle bin when app can manage media

KryptKode:
## Notes
- only proceed to delete files with `MediaStore.createDeleteRequest`
if the conversion of the FileDirItems to URI works for all the files in the list
delete casually otherwise, which would loop through all files one by one

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
@@ -111,7 +111,7 @@ fun Activity.appLaunched(appId: String) {
     }
 
     if (baseConfig.appRunCount % 40 == 0 && !baseConfig.wasAppRated) {
-        if (!resources.getBoolean(R.bool.hide_google_relations)) {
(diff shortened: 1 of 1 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
```diff
@@ -727,36 +726,45 @@ fun BaseSimpleActivity.deleteFilesBg(files: List<FileDirItem>, allowDeleteFolder
 
             if (canManageMedia()) {
                 val fileUris = getFileUrisFromFileDirItems(files).second
-                deleteSDK30Uris(fileUris) { success ->
-                    runOnUiThread {
-                        callback?.invoke(success)
+                if(fileUris.size == files.size){
```
tibbi, line 729: seems like a quite tricky check, have you tested if it works properly with SD cards too etc? I thought that you will just add a recycle bin check itself. And please format the code style with ctrl + alt + l

tibbi: will do some tests and see

KryptKode: Fixed in #1427

PR #1443, opened 2022-06-22
fix: showing multiple toast error during renaming

KryptKode:
# Notes
- remove toast error for unknown error during renaming when `success=false` to avoid multiple toasts
- show the unknown error just for cases where applicable


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
```diff
@@ -1105,6 +1106,7 @@ private fun BaseSimpleActivity.renameCasually(
                             val tempDestination = try {
                                 createTempFile(File(sourceFile.path)) ?: return@updateSDK30Uris
                             } catch (exception: Exception) {
+                                toast(R.string.unknown_error_occurred)
```
tibbi, line 1109: lets show the actual exception here with showErrorToast, just to get more info

KryptKode: I see you fixed it. Thanks

PR #1445, opened 2022-06-26
fix querying latest media

KryptKode:
# Notes
- add support for `queryArgs` `Bundle` on SDK 30+ (Android 11+) in the `queryCursorDesc` method
- found it broken in the Simple Camera app to display the last taken photo when the app starts,
-  it is also used in the Simple Gallery app to determine whether to load the directories as an optimisation

# Testing
-  using the query args bundle only works well on SDK 30+. 
- the test details are [here](https://lumpy-scorpion-7f2.notion.site/Testing-Limit-Query-ba0bbdd3ccc741c9940fefeaf43b83c5)

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
@@ -124,7 +124,7 @@ private fun Context.queryCursorDesc(
     sortColumn:String,
     limit: Int,
 ): Cursor? {
-    return if (isOreoPlus()) {
+    return if (isRPlus()) {
         val queryArgs = bundleOf(
             ContentResolver.QUERY_ARG_LIMIT to limit,
             ContentResolver.QUERY_ARG_SORT_DIRECTION to ContentResolver.QUERY_SORT_DIRECTION_DESCENDING,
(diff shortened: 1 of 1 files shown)


tibbi:
are you sure the limiting works? We had to do it like https://github.com/SimpleMobileTools/Simple-Dialer/pull/379 for the dialers recent calls

KryptKode:
> are you sure the limiting works? We had to do it like [SimpleMobileTools/Simple-Dialer#379](https://github.com/SimpleMobileTools/Simple-Dialer/pull/379) for the dialers recent calls

That method only works for Call logs as far as I know. I tested and the fix works on Android 11 and Android 9. 
Still testing at the moment with the Gallery App. 
Do you know if other apps use the `getLatestMediaId` and `getLatestMediaByDateId`  methods?

KryptKode:
Converting to draft until fully tested.

tibbi:
there are no other apps using it

KryptKode:
Just tested again and found that using the query args bundle only works well on SDK 30+. 
On lower android versions, the cursor count becomes more than 1. 
So I adjusted the if statement to use `isRPlus`. I will publish the test details soon

tibbi:
thats a long report. So can I merge this PR or you want to add anything else to it?

KryptKode:
Yes, you can merge

PR #1450, opened 2022-06-29
enter to accept and dismiss

DoozyDoz:
This is in response to [issue](https://github.com/SimpleMobileTools/Simple-Calendar/issues/1755)


dumblob:
Wow you are quick! Actually the looks OK to me. Thanks!

I am though still surprised Android API does not offer something slightly more high level to avoid such rather low-level handling (I thought there will be some flag or argument to the dialogue/form/widget constructor turning on "confirm by Enter" or alike).

Apparently I was wrong :wink:.

tibbi:
cant test it now, guess it will be ok :) Thanks

PR #1483, opened 2022-08-17
Bump dependencies

PatrykMis:


PatrykMis:
Also deprecated kotlin-android-extensions should be replaced by viewBinding: https://developer.android.com/topic/libraries/view-binding/migration#groovy

tibbi:
everything on android is deprecated, I dont want to rewrite it just for the sake of it

PR #1486, opened 2022-08-20
Extend material theme instead of appcompat

Naveen3Singh:


commons/src/main/res/values/styles.xml
```diff
@@ -228,8 +228,9 @@
         <item name="android:buttonBarNeutralButtonStyle">@style/MaterialDialogButtonStyle</item>
     </style>
 
-    <style name="Theme.Transparent" parent="Theme.AppCompat.Light.NoActionBar">
+    <style name="Theme.Transparent" parent="Theme.Material3.DayNight.NoActionBar">
```
tibbi, line 231: we arent using DayNight anywhere, do you have any specific reason why isnt it Dark?

Naveen3Singh: material widgets like the dialog use the surface color from the parent theme. If we don't use DayNight or define light/dark variants, then it'll stay dark even when using the light theme:

tibbi: we have control over the dialogs, lets use Dark theme and color them dynamically, just like always

Naveen3Singh: [done :+1: ](https://github.com/SimpleMobileTools/Simple-Commons/pull/1486/commits/89edebbc7ba252ec4d8c92b347ee14b261ef399f)

had to at least specify the dialog theme so as to not tinker with dialog creation code ([it will otherwise use theme background](https://github.com/SimpleMobileTools/Simple-Commons/blob/59d56ae16f1afe0cb13c134a42bd12b364ff8be7/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context-styling.kt#L23))

tibbi: have you tested it on lower OS versions too? It might crash because of missing material design stuff.

Naveen3Singh: I only tested on lollipop (22), 11, and 12, testing on other versions...

Naveen3Singh: all good :+1:

Naveen3Singh: > It might crash because of missing material design stuff

the "System default" option is hidden on previous versions

tibbi: Theme.Transparent can be used anywhere, it doesnt have to be system default

Naveen3Singh: Actually, Material3 is supported on older versions too (except dynamic theming obviously)

Lollipop (22)   

> It might crash because of missing material design stuff

Can confirm, it works fine.

PR #1523, opened 2022-10-25
Added Punjabi language name strings

bgo-eiu:
I am in the process of translating the Simple Mobile Tools to Punjabi, and intend to add keyboard layouts to the Simple Keyboard App. As I understand it, the language name options come from here.

Punjabi (Gurmukhi) corresponds to pa or pa-IN and Punjabi (Shahmukhi) corresponds to pa-PK. These are the two writing systems used for the language (similar situation to Chinese).

### DIFF
--- a/commons/src/main/res/values/strings.xml
+++ b/commons/src/main/res/values/strings.xml
@@ -956,6 +956,8 @@
     <string name="translation_dutch">Dutch</string>
     <string name="translation_polish">Polish</string>
     <string name="translation_portuguese">Portuguese</string>
+    <string name="translation_punjabi_in">Punjabi (Gurmukhi)</string>
+    <string name="translation_punjabi_pk">Punjabi (Shahmukhi)</string>
     <string name="translation_romanian">Romanian</string>
     <string name="translation_russian">Russian</string>
     <string name="translation_slovak">Slovak</string>


tibbi:
ah cmon, I dont think we need 2 variants of that language

bgo-eiu:
ਪੰਜਾਬੀ ਵਿੱਚ ਦੋ ਲਿਪੀ ਵਰਤਦੇ

پنجابی وچ دو لپی ورتدے

The two variants use completely different writing systems, with different alphabets and text directions. It is also the 10th most spoken language in the world.

I do need both of these variants as a good open source Android keyboard layout for either of these currently does not exist. In Pakistan the vast majority of internet users use Android and Punjabi is the most spoken language, so there are lots of people who could benefit from this. If adding these is too much I can fork it and release a separate version, but this would just make things confusing. In the Punjabi chat rooms I am in, people type in latin keyboard letters that are hard to read (no correspondance to the sounds) because they don't have a good keyboard layout.

bgo-eiu:
It is not a rare language, it is widely spoken as I noted above. There are about 21 million speakers of Serbo-Croatian (Serbian and Croatian being variants of a language), and Punjabi has about 150 million speakers. If there can only be one variant for Punjabi, then both Serbian and Croatian should be removed first, as well as Chinese (Hong Kong) (7.5 million people there).

bgo-eiu:
In any case, I do not really understand the logic here. If you do not want people to use the applications, I can maintain a fork, but I am not going to just pick one because I intend to use both.

bgo-eiu:
OK I explained that this is not a rare language, can this be merged now?

tibbi:
I dont plan adding 2 variants of that language, we will end up with hundreads of languages like that

PR #1526, opened 2022-10-27
Request Caller ID app role on android 10 and above

Naveen3Singh:


commons/src/main/res/values-ar/strings.xml
```diff
@@ -108,6 +108,7 @@
     <string name="block_unknown_calls">Block calls from not stored contacts</string>
     <string name="block_unknown_messages">Block messages from not stored contacts</string>
     <string name="add_blocked_number_helper_text">Enter a number or a pattern (e.g. *12345*, +1*8888) to block all calls and messages from numbers matching the pattern.</string>
+    <string name="must_make_default_caller_id_app">You have to make this app the default caller id app to block unknown numbers.</string>
```
tibbi, line 111: lets change it to look like at the dialog, "caller ID & spam app"

Naveen3Singh: done :+1:

PR #1536, opened 2022-11-02
Add VcfExporter and some reusable Contact classes from Simple contacts

Naveen3Singh:


tibbi:
any changes here or you just copypasted from Contacts?

Naveen3Singh:
No changes other than minor formatting, this way these can be reused in Simple contacts too

tibbi:
ok, should be good then :)

PR #1549, opened 2022-11-10
Add a bottom sheet chooser with simple list items

Naveen3Singh:


tibbi:
thanks, should be good

PR #1550, opened 2022-11-11
Bottom sheet related improvements

Naveen3Singh:
Background, theming, and title are now handled by the base bottom sheet class. 
Overriding classes just have to inflate and add whatever view they want in the `setupContentView(parent: ViewGroupt)` function.

`BottomSheetChooserDialog`makes use of this base class. 

 

(The SimpleListItemAdapter is still kept as it can be reused anywhere we need to display icon+text or a simple text-based list)


gradle.properties
```diff
@@ -11,7 +11,7 @@
 # The setting is particularly useful for tweaking memory settings.
 android.enableJetifier=true
 android.useAndroidX=true
-
+kotlin.daemon.jvmargs=-Xmx8192M
```
tibbi, line 14: lets remove this, let everyone set his own values in system

commons/src/main/res/drawable/ic_telegram_vector.xml
```diff
@@ -1,5 +1,4 @@
...
+<vector android:height="24dp" android:viewportHeight="500" android:viewportWidth="500" android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="#34aadf" android:pathData="M250 500c138.071 0 250-111.929 250-250S388.071 0 250 0 0 111.929 0 250s111.929 250 250 250z"/>
+    <path android:fillColor="#fff" android:pathData="M104.047 247.832s125-51.3 168.352-69.364c16.619-7.225 72.977-30.347 72.977-30.347s26.012-10.115 23.844 14.451c-0.723 10.116-6.503 45.52-12.283 83.815-8.671 54.191-18.064 113.439-18.064 113.439s-1.445 16.619-13.728 19.509-32.515-10.115-36.127-13.006c-2.891-2.167-54.191-34.682-72.977-50.578-5.058-4.335-10.838-13.005 0.722-23.121 26.012-23.844 57.081-53.468 75.867-72.254 8.671-8.671 17.341-28.902-18.786-4.336-51.3 35.405-101.878 68.642-101.878 68.642s-11.561 7.225-33.237 0.722c-21.677-6.502-46.966-15.173-46.966-15.173s-17.34-10.838 12.284-22.399z"/>
```
tibbi, line 3: lets always explicitly set FF as alpha

PR #1598, opened 2023-01-15
Update status bar height using WindowInsetCompat API

Naveen3Singh:
Closes https://github.com/SimpleMobileTools/Simple-SMS-Messenger/issues/561 across all affected apps.

I could not reproduce it on any other device, only one running Android 12.

Before we used the following:
```kotlin
val resourceId = resources.getIdentifier("status_bar_height", "dimen", "android")
if (resourceId > 0) {
    statusBarHeight = resources.getDimensionPixelSize(resourceId)
}
```
but evidently, that's no longer reliable. The `onApplyWindowInsets` function is an extension because it can be used by other apps (e.g. ThreadActivity in Simple-SMS-Messenger)

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt
@@ -36,6 +36,7 @@ import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.app.ActivityCompat
 import androidx.core.util.Pair
 import androidx.core.view.ScrollingView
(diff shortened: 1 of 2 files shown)


tibbi:
Im always afraid of such changes as they can have device specific glitches, but lets see. Thanks

tibbi:
this broke scrolling effects in settings, About pages etc. The navigation bar is fully transparent and the top status bar doesnt change colors.

Naveen3Singh:
working on it :+1:

Naveen3Singh:
Shall I create a new PR for the fix?

tibbi:
ye you can, but make sure it is reliable

PR #1599, opened 2023-01-16
Update status bar height using window inset API

Naveen3Singh:
Closes https://github.com/SimpleMobileTools/Simple-SMS-Messenger/issues/561 across all affected apps and doesn't have side effects like the previous (https://github.com/SimpleMobileTools/Simple-Commons/pull/1598) fix (compare the extension function)

~I could not reproduce it on any other device, only one running Android 12.~
I was able to reproduce it on devices that don't use the default status bar height (e.g. devices with a bigger notch). It also happens if you change the camera cutout in developer options as that changes the status bar height. This issue happens with the navigation bar too on devices with unique configurations hence we are updating the scroll view padding as well.

 The `onApplyWindowInsets` function is an extension because it can be used by other apps (e.g. ThreadActivity in Simple-SMS-Messenger)

**Before vs After:**

 

_tested on all the devices I could find/create_


tibbi:
did a quick test and it seems to work well so far, will see on some feedback :) Thanks

PR #1625, opened 2023-02-17
Fixed checking blocked numbers

Aga-C:
In Simple SMS Messenger, I had a situation that few conversations were missing. 

I've found out, that function checking for blocked numbers wasn't working as it should, e.g. (random example, for privacy reasons) I had blocked number `600600491`, and it was blocking conversation with number `491`.

Fix was to change string comparison from contains to equality. However, there is also the trim function and it also had to be changed. It's because, when you're trimming a number like `+48600600491` to `600600491` it's perfectly fine, but numbers can also be "SPAMMING CORP", and trimming it to "MING CORP" would lead to not blocking the conversation. That's why I decided to check by regex if the number is the real number, not just a name. I assume, that real numbers contains only numbers, `-`, `(`, `)`, `*` and `#`.

I've checked the code in Simple SMS Messenger, and it seems to work fine with both edge cases I've provided.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
@@ -1010,7 +1010,7 @@ fun Context.isNumberBlocked(number: String, blockedNumbers: ArrayList<BlockedNum
     }
 
     val numberToCompare = number.trimToComparableNumber()
-    return blockedNumbers.any { numberToCompare in it.numberToCompare || numberToCompare in it.number } || isNumberBlockedByPattern(number, blockedNumbers)
+    return blockedNumbers.any { numberToCompare == it.numberToCompare || numberToCompare == it.number } || isNumberBlockedByPattern(number, blockedNumbers)
 }
 
 fun Context.isNumberBlockedByPattern(number: String, blockedNumbers: ArrayList<BlockedNumber> = getBlockedNumbers()): Boolean {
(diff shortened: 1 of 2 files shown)


tibbi:
@Naveen3Singh please check if it didnt break anything, you were adding number blocking

Naveen3Singh:
That's a valid bug. The fix doesn't break any usual cases that I could test (numbers, formatted numbers, shortcodes, pattern blocking, etc.)

>  I decided to check by regex if the number is the real number, not just a name

We should apply this check at the following line too to disable normalization:
https://github.com/SimpleMobileTools/Simple-Commons/blob/52fd420d35273c1efb886ff541f2d57064a7369f/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt#L991

In very very rare cases, without this change, a blocked shortcode could also block a number that corresponds to the normalized version of the shortcode...

example: Blocking "GOOGLE-INC" could cause the app to block messages from "466453462" as well.

...because the normalize function converts keypad letters to numbers. Like occurrences of a, b, and c are converted to 2 and e, d, and f are converted to 3, and so on...

Aga-C:
> We should apply this check at the following line too to disable normalization:

Should I add this check, or are you going to do this?

Naveen3Singh:
> Should I add this check

sure.

Aga-C:
I've added the check to this line. Can you test if the case you've described works properly?

Naveen3Singh:
Looks good but I can't test it with physical devices and my emulator is down for some reason, I will test it later. The check itself won't break anything so this can be merged.

tibbi:
ok, thanks guys

PR #1637, opened 2023-03-07
fix: remove showTabs pref from commons

ismailnurudeen:
This preference should live in the individual config file of each app.


tibbi:
Simple Dialer doesnt have showtabs at all, fix it please

PR #1639, opened 2023-03-12
Delete hidden files using casual method

Naveen3Singh:
Closes https://github.com/SimpleMobileTools/Simple-Gallery/issues/2759

This is because the usual `MediaStore.createDeleteRequest()` doesn't seem to work for files that are hidden. 

**Note:** If a folder's `.nomedia` file is deleted i.e. unhidden, deleting any media inside the folder will still fail if MediaStore isn't notified about the `.nomedia` deletion. To fix that, we must properly update MediaStore when we hide/unhide a folder.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
@@ -702,19 +702,19 @@ fun BaseSimpleActivity.deleteFilesBg(files: List<FileDirItem>, allowDeleteFolder
     }
 
     val firstFile = files.first()
-    handleSAFDialog(firstFile.path) {
+    val firstFilePath = firstFile.path
+    handleSAFDialog(firstFilePath) {
         if (!it) {
             return@handleSAFDialog
         }
 
-        checkManageMediaOrHandleSAFDialogSdk30(firstFile.path) {
+        checkManageMediaOrHandleSAFDialogSdk30(firstFilePath) {
             if (!it) {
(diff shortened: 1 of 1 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
```diff
@@ -713,7 +713,8 @@ fun BaseSimpleActivity.deleteFilesBg(files: List<FileDirItem>, allowDeleteFolder
             }
 
             val recycleBinPath = firstFile.isRecycleBinPath(this)
-            if (canManageMedia() && !recycleBinPath) {
+            val containsNoMedia = firstFile.path.doesThisOrParentHaveNoMedia(HashMap(), null)
+            if (canManageMedia() && !recycleBinPath && !containsNoMedia) {
```
tibbi, line 717: Lets avoid calling containsNoMedia when it isnt necessary, it can often be slow. I guess the best way (not the way we do it mostly) is just removing the local variable and changing it to
if (canManageMedia() && !recycleBinPath && !firstFile.path.doesThisOrParentHaveNoMedia(HashMap(), null)) {

It could be solved other ways too, but there is an else statement too so it could be messy.

Naveen3Singh: done :+1:

Naveen3Singh:
https://github.com/SimpleMobileTools/Simple-Gallery/pull/2800 should be merged too. Otherwise, the gallery won't update the media store properly when a folder is unhidden (thus it'll fail to delete its content)

PR #1648, opened 2023-03-27
Truncate all output files

t-m-w:
Address the behavior change from [Android bug 180526528](https://issuetracker.google.com/issues/180526528) so that outputting to a URI that already exists replaces the entire destination instead of only modifying part of it, when the output is smaller.

This resolves an issue with Simple Gallery in which resizing a file to a smaller size when that app only has Media Management permission (as opposed to All Files permission) leaves extra data at the end, with the file not getting any smaller.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
@@ -1208,7 +1208,7 @@ fun BaseSimpleActivity.getFileOutputStream(fileDirItem: FileDirItem, allowCreati
                 if (!getDoesFilePathExist(fileDirItem.path)) {
                     createAndroidSAFFile(fileDirItem.path)
                 }
-                callback.invoke(applicationContext.contentResolver.openOutputStream(uri))
+                callback.invoke(applicationContext.contentResolver.openOutputStream(uri, "wt"))
             }
         }
         needsStupidWritePermissions(fileDirItem.path) -> {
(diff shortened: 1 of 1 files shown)


t-m-w:
P.S. I am not sure how to reproduce any issues with Google Drive or if they are still a problem, but if concerned, could possibly use "rwt" instead of "wt", as was done for this:
- https://github.com/SimpleMobileTools/Simple-Notes/issues/444
- https://github.com/SimpleMobileTools/Simple-Notes/pull/445

Let me know if you want me to change it to rwt.

tibbi:
have you tried if it doesnt break anything?

t-m-w:
> have you tried if it doesnt break anything?

I'm not seeing any breakage on an Android 13 device or in a Play emulator for Android 10 and 11. I tried resizing files on top of themselves and to other destinations. I tried building Simple Draw with this library change and saving an SVG directly into Google Drive or other locations, Android 10 and 11, and everything saved fine, but Simple Draw still does not truncate files saved on top of each other, so the problem is not fixed absolutely everywhere.

Separately, I also tried building Simple Notes modified to use "wt" with openOutputStream instead of "rwt" (undoing patch linked earlier) and tested exporting to Google Drive, both on Android 10 and 11, just to check if "wt" is still a problem for Google Drive, and it doesn't look like it.

tibbi:
@Naveen3Singh can you roughly test if this breaks anything?

Naveen3Singh:
It works. In the file size update issue, `FileOutputStream` worked because it still uses the `O_TRUNC` flag by default. This PR is the proper fix for the file size update issue in Simple Gallery (the previous changes I did can be reverted once this is merged) and other apps.

> I am not sure how to reproduce any issues with Google Drive or if they are still a problem

Google recommends using `rwt` but that was in 2021, they probably added support for `wt` later.
 
> Simple Draw still does not truncate files saved on top of each other, so the problem is not fixed absolutely everywhere.

It's not clear why Simple Draw isn't able to truncate the files, more research is needed.

PR #1649, opened 2023-03-28
Simple-Contacts: Added support for relations

rlieger:
A ContactRelation can be used to express a relation to another person (e.g. spouse, friend, colleague) within a contact. This concept is supported by both Android (ContactContract.Relation) and vCard 4.0 (RELATED) and should thus also be supported in SimpleContacts.
Note: Like most fields, displaying relations for a contact is optional and can be enabled/disabled in the manage-visable-fields dialog. It is disabled by default, so to see my code in action be sure to enable it...

I have added suitable code for both the base types (in Simple-Commons) and the application logic for the viewing/editing the databases (in Simple-Contacts). Thus to obtain a working system, pulling both Simple-Commons and Simple-Contacts is necessary!


rlieger:
This pull request was not handled for three weeks. I have since worked a lot on the code and hope to release an improved version within a week or two. I suggest that you skip this version and directly integrate the new version.

Naveen3Singh:
> hope to release an improved version within a week or two

then I guess this one can be closed?

PR #1651, opened 2023-03-31
Fixed repeating first name into surname

Aga-C:
Fixes https://github.com/SimpleMobileTools/Simple-Dialer/issues/570


commons/src/main/kotlin/com/simplemobiletools/commons/helpers/MyContactsContentProvider.kt
```diff
@@ -88,7 +88,7 @@ class MyContactsContentProvider {
                             val anniversaries = Gson().fromJson<ArrayList<String>>(anniversariesJson, stringsToken) ?: ArrayList()
                             val names = name.split(" ")
                             val firstName = names.firstOrNull() ?: ""
-                            val surname = names.lastOrNull() ?: ""
+                            val surname = if (names.size > 1) names.lastOrNull() ?: "" else ""
```
tibbi, line 91: add some brackets and split it into multiple lines please to make it more readable

tibbi:
should be good now, thanks

PR #1657, opened 2023-04-10
Update strings.xml

spkprs:


tibbi:
updated the strings manually due to some messup. Thanks

PR #1663, opened 2023-04-18
feat: added method to get primary number

ismailnurudeen:


commons/src/main/kotlin/com/simplemobiletools/commons/models/contacts/Contact.kt
```diff
@@ -237,4 +237,9 @@ data class Contact(
     fun isPrivate() = source == SMT_PRIVATE
 
     fun getSignatureKey() = if (photoUri.isNotEmpty()) photoUri else hashCode()
+    fun getPrimaryNumber(): String? {
+        val numbers = phoneNumbers
```
tibbi, line 241: isnt this numbers variable unnecessary? Also add an empty line before new functions

PR #1667, opened 2023-04-25
Apply IME insets when using transparent navigation

Naveen3Singh:
This makes sure that the views are always above the keyboard. No side effects so far.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/activities/BaseSimpleActivity.kt
@@ -228,22 +228,22 @@ abstract class BaseSimpleActivity : AppCompatActivity() {
             if (navigationBarHeight > 0 || isUsingGestureNavigation()) {
                 window.decorView.systemUiVisibility = window.decorView.systemUiVisibility.addBit(View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION)
                 updateTopBottomInsets(statusBarHeight, navigationBarHeight)
-                onApplyWindowInsets {
-                    val insets = it.getInsets(WindowInsetsCompat.Type.systemBars())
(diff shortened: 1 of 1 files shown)


Naveen3Singh:
The only issue with this solution is that the scroll view doesn't scroll up automatically to the top when a field from the bottom is selected for input.

The alternative is disabling the transparent navigation bar in activities requiring keyboard input. For e.g: `EventActivity` in Simple Calendar has the following call:
```kotlin
updateMaterialActivityViews(event_coordinator, event_holder, useTransparentNavigation = true, useTopSearchMenu = false)
```

and `adjustResize` is broken as a result of using transparent navigation (which by the way doesn't seem to be working on newer versions)

tibbi:
Ive always been afraid of such changes, they often break something on some device. Will have to test it properly

Naveen3Singh:
If the glitch isn't reported in other apps and you wanna avoid adding this solution for now, we can do this in Simple Calendar only (or disable transparent navigation in event activity)

Naveen3Singh:
though I haven't found any issues due to this change on my virtual and physical devices. Tested it with some other simple apps too.

tibbi:
lets start with Simple Contacts, it has less users than the calendar

Naveen3Singh:
okay, here it is: https://github.com/SimpleMobileTools/Simple-Contacts/pull/974

tibbi:
will test it in some real app, update commons there please

Naveen3Singh:
> update commons there 

Do you mean update it in the [Simple Calender PR](https://github.com/SimpleMobileTools/Simple-Calendar/pull/2059)?

(The Simple Contact EditContactActivity is independent and doesn't use this method from commons)

tibbi:
right, ok

tibbi:
Test the latest gallery please, in landscape it applies the navigation bar width twice. I wasnt researching it, not sure if it is related to this commit.

screenshot

PR #1689, opened 2023-05-23
fix #530

liyamahendra:
Even though our Dialer app was configured to be the default app, when calling a number, the device used to prompt to select the app.

Adding this fix, where we check if our app is the default dialer or not. If it is, we explicitly open our app when calling the `launchCallIntent`.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
@@ -550,10 +550,7 @@ fun BaseSimpleActivity.launchCallIntent(recipient: String, handle: PhoneAccountH
             }
 
             if (isDefaultDialer()) {
-                val packageName = when(BuildConfig.DEBUG) {
-                    true -> "com.simplemobiletools.dialer.debug"
-                    false -> "com.simplemobiletools.dialer"
-                }
+                val packageName = if (BuildConfig.DEBUG) "com.simplemobiletools.dialer.debug" else "com.simplemobiletools.dialer"
                 val className = "com.simplemobiletools.dialer.activities.DialerActivity"
                 this.setClassName(packageName, className)
             }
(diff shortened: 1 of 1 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
```diff
@@ -537,7 +538,6 @@ fun Activity.launchViewContactIntent(uri: Uri) {
         launchActivityIntent(this)
     }
 }
-
 fun BaseSimpleActivity.launchCallIntent(recipient: String, handle: PhoneAccountHandle? = null) {
```
tibbi, line 541: lets keep empty spaces before the functions

commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
```diff
@@ -548,6 +548,15 @@ fun BaseSimpleActivity.launchCallIntent(recipient: String, handle: PhoneAccountH
                 putExtra(TelecomManager.EXTRA_PHONE_ACCOUNT_HANDLE, handle)
             }
 
+            if (isDefaultDialer()) {
+                val packageName = when(BuildConfig.DEBUG) {
```
tibbi, line 552: lets use an if/else...

liyamahendra: You mean instead of `when` use `if-else`?

liyamahendra:
@tibbi changes have been done.

commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Activity.kt
tibbi: ye, "when" with "true" and "false" is just so unprefessional and looks wrong

tibbi:
ok thanks, should be good

tibbi:
ok thanks, should be good

PR #1696, opened 2023-06-02
Cursor fix when on contact clicked

Merkost:
Removed Data.MIMETYPE from selectionArgs in contentResolver query in order to allow opening contacts if only phone number is filled


tibbi:
what exactly does this fix and how to reproduce it?

Merkost:
There was a bug, when you add a new contact with filling only the phone number field (In a google contacts app for example), then it's impossible to open it. However, I was unable to repeat it after my other pull request, where I sat the name to the contact's number, if the name wasn't typed. I'll close it for now

https://github.com/SimpleMobileTools/Simple-Commons/assets/56008710/7c314bd5-fb32-43f0-b313-afd2ba8fd156

PR #1714, opened 2023-06-23
Dialer new block hidden calls feature

Merkost:


tibbi:
could you actually reproduce and test calls from hidden numbers?

tibbi:
will test it properly in dialer itself

PR #1721, opened 2023-07-05
Fix scaling of breadcrumbs for very large text sizes

esensar:
Breadcrumbs component previously had fixed height which made it cut off text when it is changed to largest sizes in settings.

This changes breadcrumbs to have dynamic height, based on contents, with just the minimum height based on previously fixed size.

The previously fixed height was also put on items as minimum height, to ensure consistent behavior with smaller fonts too, but this will have effect only on very small font sizes (not realistic). Centering of text was added for that reason too. It can be removed if needed, since it is needed only for unrealistically small font sizes.


tibbi:
looks good, thanks

PR #1725, opened 2023-07-07
Add more controls for EnterPasswordDialog

esensar:


commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/EnterPasswordDialog.kt
```diff
@@ -39,4 +41,15 @@ class EnterPasswordDialog(
                 }
             }
     }
+
+    fun dismiss(notify: Boolean = true) {
+        if (!notify) {
+            dialog?.setOnDismissListener(null)
```
tibbi, line 47: what is this clearing about? I dont really get it.

esensar: In this case, I didn't want the `cancelCallback` to be invoked, because I would use this to dismiss the dialog when correct password is entered. If I left the callback, it would invoke `cancelCallback`, which in the case of PDF opening would finish the activity.

tibbi: ok, will see how it behaves in the app itself

PR #1738, opened 2023-07-15
Contacts grid shared files

Merkost:


commons/src/main/kotlin/com/simplemobiletools/commons/helpers/BaseConfig.kt
```diff
@@ -549,4 +549,13 @@ open class BaseConfig(val context: Context) {
...
         get() = prefs.getBoolean(FAVORITES_CUSTOM_ORDER_SELECTED, false)
         set(selected) = prefs.edit().putBoolean(FAVORITES_CUSTOM_ORDER_SELECTED, selected).apply()
+
+    var viewType: Int
+        get() = prefs.getInt(VIEW_TYPE, VIEW_TYPE_LIST)
+        set(viewType) = prefs.edit().putInt(VIEW_TYPE, viewType).apply()
+
+    var contactsGridColumnCnt: Int
+        get() = prefs.getInt(CONTACTS_GRID_COLUMN_COUNT, CONTACTS_GRID_DEFAULT_COLUMNS_COUNT)
+        set(gridLayoutSpanCount) = prefs.edit().putInt(CONTACTS_GRID_COLUMN_COUNT, gridLayoutSpanCount).apply()
```
tibbi, line 559: use the same variable names in configs on line 557 and 559. Either gridLayoutSpanCount or contactsGridColumnCnt

Merkost: Oh, completely forgot about it. Thanks.

commons/src/main/kotlin/com/simplemobiletools/commons/helpers/BaseConfig.kt
```diff
@@ -549,4 +549,13 @@ open class BaseConfig(val context: Context) {
...
         get() = prefs.getBoolean(FAVORITES_CUSTOM_ORDER_SELECTED, false)
         set(selected) = prefs.edit().putBoolean(FAVORITES_CUSTOM_ORDER_SELECTED, selected).apply()
+
+    var viewType: Int
+        get() = prefs.getInt(VIEW_TYPE, VIEW_TYPE_LIST)
+        set(viewType) = prefs.edit().putInt(VIEW_TYPE, viewType).apply()
+
+    var contactsGridColumnCnt: Int
+        get() = prefs.getInt(CONTACTS_GRID_COLUMN_COUNT, CONTACTS_GRID_DEFAULT_COLUMNS_COUNT)
+        set(gridLayoutColumnCount) = prefs.edit().putInt(CONTACTS_GRID_COLUMN_COUNT, gridLayoutColumnCount).apply()
```
tibbi, line 559: use consistent variable names, do not mix gridLayoutColumnCount and contactsGridColumnCnt

commons/src/main/kotlin/com/simplemobiletools/commons/helpers/Constants.kt
```diff
@@ -185,6 +185,13 @@ const val SHOW_PRIVATE_CONTACTS = "show_private_contacts"
 const val MERGE_DUPLICATE_CONTACTS = "merge_duplicate_contacts"
 const val FAVORITES_CONTACTS_ORDER = "favorites_contacts_order"
 const val FAVORITES_CUSTOM_ORDER_SELECTED = "favorites_custom_order_selected"
+const val VIEW_TYPE = "view_type"
+const val CONTACTS_GRID_COLUMN_COUNT = "favourites_column_count"
```
tibbi, line 189: unify constant and string here too

commons/src/main/kotlin/com/simplemobiletools/commons/helpers/Constants.kt
```diff
@@ -185,6 +185,13 @@ const val SHOW_PRIVATE_CONTACTS = "show_private_contacts"
...
+
+// contact grid view constants
+const val CONTACTS_GRID_DEFAULT_COLUMNS_COUNT = 3
```
tibbi, line 192: and this needs dynamic default values too

commons/src/main/res/values/integers.xml
```diff
@@ -3,4 +3,6 @@
     <integer name="default_sorting">1</integer>
     <integer name="default_font_size">1</integer>
     <integer name="default_viewpager_page">0</integer>
+    <integer name="contacts_grid_columns_count_portrait">3</integer>
+    <integer name="contacts_grid_columns_count_landscape">5</integer>
```
tibbi, line 7: man... the point of this is having different numbers in different resolutions. If you use only have one values/integers.xml, it makes no sense.

Merkost: But that's how it is done in launcher app

Merkost: @tibbi, anyway, what are we going to accomplish by extracting the default column count in resources?

tibbi: not true, the launcher has different values depending on the resolution. For now there are 2 variants only, but more can be added easily.

Merkost: > not true, the launcher has different values depending on the resolution. For now there are 2 variants only, but more can be added easily.

Yeah, I didn't see it in GitHub files, but after importing the project to android studio I was able to see that there are two values. I'll implement the same in commons now. Sorry for the inconvenience.

commons/src/main/kotlin/com/simplemobiletools/commons/helpers/BaseConfig.kt
```diff
@@ -549,4 +552,26 @@ open class BaseConfig(val context: Context) {
...
+                context.resources.getInteger(R.integer.contacts_grid_columns_count_portrait),
+                context.resources.getInteger(R.integer.contacts_grid_columns_count_landscape)
+            )
+        )
+        set(contactsGridColumnCnt) = prefs.edit().putInt(
+            valueByConfiguration(CONTACTS_GRID_COLUMN_COUNT_PORTRAIT, CONTACTS_GRID_COLUMN_COUNT_LANDSCAPE),
+            contactsGridColumnCnt
+        ).apply()
+
+    private fun <T> valueByConfiguration(portraitValue: T, landScapeValue: T): T {
```
tibbi, line 573: this is a really complicated solution to an easy problem.. Just check how is portraitColumnCnt and landscapeColumnCnt done in App Launcher, that is a lot easier to read.

Merkost: I have simplified the function. Could you please take a look at the implementation now?

tibbi: this should be good enough, renamed some things a bit

tibbi:
thanks, code looks ok, havent checked functionality itself yet. Will do it in a real app.

PR #1744, opened 2023-07-16
Fix build issue by updating permission required function call

wilsonrc:
- Fix the build issue by updating the permission required function call with proper arguments.
- update the open alarm settings to open a specific alarm settings page.


tibbi:
PermissionRequiredDialog is used in quite a few apps though, not only in commons. For such changes you should query all apps and see where it is used, so they can be built too.

PR #1762, opened 2023-07-24
[WIP] view binding migration

FunkyMuse:


gradle/libs.versions.toml
```diff
@@ -0,0 +1,142 @@
+[versions]
```
tibbi, line 1: this file has been somehow generated?

FunkyMuse: I write it myself, I wrote a [blogpost](https://funkymuse.dev/posts/toml-gradle/) explaining how it works.

FunkyMuse: Although it will become the norm with Android Studio Hedgehog and above

commons/src/main/kotlin/com/simplemobiletools/commons/adapters/MyRecyclerViewAdapter.kt
```diff
@@ -91,7 +91,7 @@ abstract class MyRecyclerViewAdapter(val activity: BaseSimpleActivity, val recyc
 
                 if (baseConfig.isUsingSystemTheme) {
                     actBarTextView?.onGlobalLayout {
-                        val backArrow = activity.findViewById<ImageView>(R.id.action_mode_close_button)
+                        val backArrow = activity.findViewById<ImageView>(androidx.appcompat.R.id.action_mode_close_button)
```
FunkyMuse, line 94: Could this be Material or AppCompat or another, I wasn't sure which one the synthetics was importing

commons/src/main/kotlin/com/simplemobiletools/commons/adapters/MyRecyclerViewListAdapter.kt
```diff
@@ -99,7 +99,7 @@ abstract class MyRecyclerViewListAdapter<T>(
 
                 if (baseConfig.isUsingSystemTheme) {
                     actBarTextView?.onGlobalLayout {
-                        val backArrow = activity.findViewById<ImageView>(R.id.action_mode_close_button)
+                        val backArrow = activity.findViewById<ImageView>(androidx.appcompat.R.id.action_mode_close_button)
```
FunkyMuse, line 102: Could this be Material or AppCompat or another, I wasn't sure which one the synthetics was importing

tibbi:
so I should decide which icon to use? I dont mind, just be consistent across all apps and screens.

commons/src/main/kotlin/com/simplemobiletools/commons/activities/LicenseActivity.kt
```diff
@@ -3,25 +3,69 @@ package com.simplemobiletools.commons.activities
...
+import com.simplemobiletools.commons.helpers.LICENSE_PHOTOVIEW
+import com.simplemobiletools.commons.helpers.LICENSE_PICASSO
+import com.simplemobiletools.commons.helpers.LICENSE_REPRINT
+import com.simplemobiletools.commons.helpers.LICENSE_ROBOLECTRIC
+import com.simplemobiletools.commons.helpers.LICENSE_RTL
+import com.simplemobiletools.commons.helpers.LICENSE_SANSELAN
+import com.simplemobiletools.commons.helpers.LICENSE_SMS_MMS
+import com.simplemobiletools.commons.helpers.LICENSE_STETHO
+import com.simplemobiletools.commons.helpers.LICENSE_SUBSAMPLING
+import com.simplemobiletools.commons.helpers.LICENSE_ZIP4J
```
tibbi, line 50: guess we should use an asterisk at importing these, as previously

FunkyMuse: It's a bad development practice to use asterisk import

FunkyMuse:
> so I should decide which icon to use? I dont mind, just be consistent across all apps and screens.

It's actually not what icon to use but it tries to find it from the activity (BaseSimpleActivity), I'm asking which one was used before so that we don't break compatibility so I assumed AppCompat since I think even Material inherits from AppCompat (but I might be wrong, it's been some time since I've used XML)

commons/src/main/kotlin/com/simplemobiletools/commons/activities/LicenseActivity.kt
tibbi: lets use asterisks after 5 imports, it is written in editorconfig file too

tibbi:
lets try appcompat and we will see how it looks

PR #1770, opened 2023-07-28
feat: add necessary Common compose things to expose to other apps

FunkyMuse:


commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/Theme.kt
```diff
@@ -0,0 +1,79 @@
...
+    theme: Theme = systemDefaultMaterialYou(),
+    content: @Composable () -> Unit,
+) {
+    val view = LocalView.current
+    val context = LocalContext.current
+    val systemUiController = rememberSystemUiController()
+    val baseConfig = remember { context.config }
+
+    val colorScheme = if (!view.isInEditMode) {
+
```
tibbi, line 31: avoid such empty lines please. Also having "val colorScheme" twice under each other is really confusing

commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/Theme.kt
```diff
@@ -0,0 +1,79 @@
...
+        val colorScheme = when {
+            theme is Theme.SystemDefaultMaterialYou && isSPlus() -> {
+                if (isSystemInDarkTheme()) dynamicDarkColorScheme(context) else dynamicLightColorScheme(context)
```
tibbi, line 34: split such to multiple lines please. If it crashes, I wont know which function on the line crashed.

tibbi:
should be good

PR #1771, opened 2023-07-29
add Contact Name to blocked number.

wilsonrc:
### Description

Add Contact Name to Manage Blocked Numbers Screen.

I created a new method to load the blocked numbers with the contact name to avoid breaking the other uses of the getBlockedNumber function.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
@@ -996,6 +996,47 @@ fun Context.isDefaultDialer(): Boolean {
     }
 }
 
+fun Context.getContactsHasMap(callback: (HashMap<String, String>) -> Unit) {
+    ContactsHelper(this).getContacts(showOnlyContactsWithNumbers = true) { contactList ->
+        val privateContacts: HashMap<String, String> = HashMap()
+        for (contact in contactList) {
+            for (phoneNumber in contact.phoneNumbers) {
+                privateContacts[phoneNumber.value] = contact.name
+            }
+        }
+        callback(privateContacts)
+    }
+}
+
+@TargetApi(Build.VERSION_CODES.N)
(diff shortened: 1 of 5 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
```diff
@@ -992,6 +992,47 @@ fun Context.isDefaultDialer(): Boolean {
...
+            BlockedNumbers.COLUMN_E164_NUMBER,
+        )
+
+        queryCursor(uri, projection) { cursor ->
+            val id = cursor.getLongValue(BlockedNumbers.COLUMN_ID)
+            val number = cursor.getStringValue(BlockedNumbers.COLUMN_ORIGINAL_NUMBER) ?: ""
+            val normalizedNumber = cursor.getStringValue(BlockedNumbers.COLUMN_E164_NUMBER) ?: number
+            val comparableNumber = normalizedNumber.trimToComparableNumber()
+
+            val contactName = contacts[number] ?: getString(R.string.unknown)
```
tibbi, line 1028: does this also handle settings like Start name with surname properly?

wilsonrc: I think So. This is using the name property  `Contact.name` that uses the `getNameToDisplay()` which internally handles the start name with surname preference.

wilsonrc: I think So. This is using the name property  `Contact.name` that uses the `getNameToDisplay()`, which internally handles the start name with surname preference.

tibbi:
test it please, dont just think. Anyway, updating commons is a bit more difficult now, you have to update target SDK to 34 of the given app too, ideally rewrite to viewbinding aswell.

commons/src/main/kotlin/com/simplemobiletools/commons/extensions/Context.kt
wilsonrc: @tibbi I tested it, and it handles start with surname correctly.

PR #1777, opened 2023-08-03
feat: migration to Compose

FunkyMuse:
also introduced two new components for building settings looking scaffold and break down components to be reusable

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/Theme.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/Theme.kt
@@ -64,13 +64,12 @@ internal fun Theme(
     }
 
     SideEffect {
-        systemUiController.setNavigationBarColor(Color.Transparent.copy(alpha = 0.1f), darkIcons = !isSystemInDarkTheme) //https://github.com/google/accompanist/issues/1699
+        systemUiController.setNavigationBarColor(Color.Transparent.copy(alpha = 0.25f), darkIcons = !isSystemInDarkTheme)    // https://github.com/google/accompanist/issues/1699
     }
 
     SideEffect {
         updateRecentsAppIcon(baseConfig, context)
     }
-
 
     MaterialTheme(
         colorScheme = colorScheme,


commons/src/main/kotlin/com/simplemobiletools/commons/compose/settings/scaffold/SettingsScaffold.kt
```diff
@@ -0,0 +1,77 @@
...
+                    .verticalScroll(rememberScrollState())
+            ) {
+                content()
+            }
+        }
+    }
+}
+
+
+
```
tibbi, line 65: man, look for those double and triple empty lines...

tibbi:
lets use smaller line height at questions, the empty gap between 2 lines of long questions is too big. And we should add a divider after each answer

commons/src/main/kotlin/com/simplemobiletools/commons/compose/settings/scaffold/SettingsScaffold.kt
FunkyMuse: Time to add Spotless for the formatting 🫠

FunkyMuse:
> lets use smaller line height at questions, the empty gap between 2 lines of long questions is too big. And we should add a divider after each answer

let me know how it looks like now

tibbi:
line height still hasnt changed at questions, they are too high. And add some more spacing below and under the divider

tibbi:
and lets try writing the questions with same text size as the answers

FunkyMuse:
> and lets try writing the questions with same text size as the answers

please check now

tibbi:
it needs some more spacing. Lets try adding more left margin, it is too close to the screens side. And increase margins under and above divider

FunkyMuse:
> it needs some more spacing. Lets try adding more left margin, it is too close to the screens side. And increase margins under and above divider

The left margin is the one from the default ListItem from M3 Compose, i added some more left margin and increased the margins under and above the divider, do they need to be equal top/bottom for the divider or top can be smaller and bottom higher?

tibbi:
well, it looks better with 10dp bottom padding instead of 4, but I still dont think it is the most modern solution to listing faq items like that :/ What other options do we have?

FunkyMuse:
> well, it looks better with 10dp bottom padding instead of 4, but I still dont think it is the most modern solution to listing faq items like that :/ What other options do we have?

yeah I agree, doesn't look good

Maybe something like this https://dribbble.com/shots/14043052-Daily-UI-092-F-A-Q?

tibbi:
thats looks better, but Id like to have all items open for reading quickly through them

commons/src/main/kotlin/com/simplemobiletools/commons/compose/screens/ContributorsScreen.kt
```diff
@@ -0,0 +1,148 @@
...
+@MyDevices
+private fun ContributorsScreenPreview() {
+    AppThemeSurface {
+        ContributorsScreen(
+            goBack = {},
+            contributors = listOf(
+                LanguageContributor(R.drawable.ic_flag_arabic_vector, R.string.translation_arabic, R.string.translators_arabic),
+                LanguageContributor(R.drawable.ic_flag_azerbaijani_vector, R.string.translation_azerbaijani, R.string.translators_azerbaijani),
+                LanguageContributor(R.drawable.ic_flag_bengali_vector, R.string.translation_bengali, R.string.translators_bengali),
+                LanguageContributor(R.drawable.ic_flag_catalan_vector, R.string.translation_catalan, R.string.translators_catalan),
```
tibbi, line 143: wheres the rest?

commons/src/main/kotlin/com/simplemobiletools/commons/compose/screens/LicenseScreen.kt
```diff
@@ -0,0 +1,117 @@
...
+                License(LICENSE_ESPRESSO, R.string.espresso_title, R.string.espresso_text, R.string.espresso_url),
+                License(LICENSE_GSON, R.string.gson_title, R.string.gson_text, R.string.gson_url),
+                License(LICENSE_LEAK_CANARY, R.string.leak_canary_title, R.string.leakcanary_text, R.string.leakcanary_url),
+                License(LICENSE_NUMBER_PICKER, R.string.number_picker_title, R.string.number_picker_text, R.string.number_picker_url),
+                License(LICENSE_EXOPLAYER, R.string.exoplayer_title, R.string.exoplayer_text, R.string.exoplayer_url),
+                License(LICENSE_PANORAMA_VIEW, R.string.panorama_view_title, R.string.panorama_view_text, R.string.panorama_view_url),
+                License(LICENSE_SANSELAN, R.string.sanselan_title, R.string.sanselan_text, R.string.sanselan_url),
+                License(LICENSE_FILTERS, R.string.filters_title, R.string.filters_text, R.string.filters_url),
+                License(LICENSE_GESTURE_VIEWS, R.string.gesture_views_title, R.string.gesture_views_text, R.string.gesture_views_url),
+                License(
```
tibbi, line 95: this didnt fit into 1 line?

tibbi:
issues:
1. bottom navigation bar is not transparent
2. too big font size at Contributors -> Development names
3. FAQ titles use too bold text and divider lines dont cross whole width. Use same style like at licenses, everything is good there

commons/src/main/kotlin/com/simplemobiletools/commons/compose/screens/ContributorsScreen.kt
FunkyMuse: I just copy pasted the same list that was there, am I missing something?

FunkyMuse: Although there's this logic

` val licenseMask = remember { intent.getLongExtra(APP_LICENSES, 0) or LICENSE_KOTLIN }
            val thirdPartyLicenses = remember { initLicenses().filter { licenseMask and it.id != 0L }.toImmutableList() }`

which only displays 2 of them if not launched with an intent extra

tibbi:
right, it is an important part of the UI though, how do other apps usually solve it? I thought it is one of the most basic things.

tibbi:
it should be just semtransparent, not fully. Sorry for bad wording. But if I scroll Contributors all the way down, the bottom text is behind the navigation bar. There should be some margin to make sure it is readable.

FunkyMuse:
> it should be a bit more transparent. I dont really have a specific color code, not sure how was it set previously..

I changed the alpha now, i think that's it, as i compared to other apps, i just need to know if i should bump it up more or this is enough

tibbi:
Licenses and Contributor names arent using proper text color

tibbi:
as I keep saying, test all UI changes in different themes please...

FunkyMuse:
> Licenses and Contributor names arent using proper text color

what do you mean by that?

I thought that this time around there wouldn't be hacky things for the color, it's using the primary one from the theme or there's some different thing that i didn't quite catch?

PR #1780, opened 2023-08-04
Add helpers and strings for full screen notification permission

esensar:


commons/src/main/res/values-ar/strings.xml
```diff
@@ -178,10 +178,12 @@
     <string name="no_audio_permissions">مطلوب إذن الصوت</string>
     <string name="no_permission">لاتوجد صلاحيات</string>
     <string name="allow_notifications_reminders">يجب أن تسمح للتطبيقات لعرض الإخطارات ، أو أنها لن تكون قادرة على عرض تذكير .</string>
+    <string name="allow_full_screen_notifications_reminders">You should allow the app to display full screen notifications, else you may miss some reminders.</string>
```
tibbi, line 181: lets change "the app to display" to "the app displaying" I think

tibbi:
these permissions are actually mandatory, right? Because other strings are like "You must", while now you wrote "You should". If we cannot show incoming calls while the screen is off without this permission, it should be mandatory. Same with event notifications. If the users wont see reminders without that permission while the device is off, they will report it as bugs, so we should make it mandatory.

esensar:
I went for that string because we can still show the notification, but I think you are right. The won't see it if the screen is off, so we should probably make it mandatory then :/
Let me update the strings then

tibbi:
ye looks good, thanks. Didnt know the new OS is called Upside Down Cake, thats a pretty bad name ;D

esensar:
> ye looks good, thanks. Didnt know the new OS is called Upside Down Cake, thats a pretty bad name ;D

Yeah, me neither, only saw it when I was adding these helpers :smile:

PR #1786, opened 2023-08-16
fix: proper text color for contributors and licenses

FunkyMuse:


tibbi:
thanks, should be good

tibbi:
It still misbehaves a bit in landscape mode, when the navigation bar is on the side. The navigation bar shouldnt be transparent there and nothing should show behind it.
And in light themes the navigation bar should be translucent-white, not translucent-black
Just check some old apps to see what is the intended behaviour.

FunkyMuse:
> It still misbehaves a bit in landscape mode, when the navigation bar is on the side. The navigation bar shouldnt be transparent there and nothing should show behind it.

Are we talking about tablets here?

tibbi:
no, casual phones

PR #1787, opened 2023-08-17
fix: navigation bar color on dark/light background

FunkyMuse:
@tibbi let me know if the alpha is okay for the white backgrounds


tibbi:
looks good, thanks

PR #1788, opened 2023-08-17
fix: navigation bar paddings for landscape mode

FunkyMuse:


tibbi:
I think this should be solved in some generic way so that we dont have to think about it in all apps screens. The top actionmenu color shouldnt be behind the navigation either, just like the divider bar. Check Simple Clock, while the Settings page works well, the About doesnt.

FunkyMuse:
> 

i think my latest change fixes this, accordingly as at Simple Clock

tibbi:
should be good, will do more testing in a real app. Thanks

tibbi:
if the navigation bar is white, buttons Back, Home and Recents should be dark

tibbi:
and in landscape mode the navigation is always dark, even when it should be light

FunkyMuse:
> if the navigation bar is white, buttons Back, Home and Recents should be dark

Which should be the source color for all?

tibbi:
what source color you mean? the navigation bar itself properly toggles light/dark color, so just the buttons have to be adjusted too

FunkyMuse:
> what source color you mean? the navigation bar itself properly toggles light/dark color, so just the buttons have to be adjusted too

Oh you meant the navigation bar's (3 buttons) colors, alright, gotcha.

PR #1792, opened 2023-08-22
Add string for allowing access to more media

esensar:


tibbi:
this requires some new permission? Havent tested it myself yet, might need a better string.

esensar:
> this requires some new permission? Haven't tested it myself yet, might need a better string.

Yeah. The applications will work fine without the new permission too, but the UX won't be ideal (the app will keep asking users to give permission, unless they give permission to all media right away). This item is meant to avoid nagging users about that permission and let them manually add access to more media.

tibbi:
oh I see the explanation at that other PR. Will have to research how do other galleries work on Android 14, this looks pretty weird :)

esensar:
I am not sure if there is a way to always require users to give access to all media, since partial permission doesn't really make much sense for a gallery application :/

tibbi:
this new string will still be used in gallery only, or no?

esensar:
I think so. I don't think other apps use images or video. I know that SMS uses them, but using a different API which doesn't need this permission.

tibbi:
accepting it, but I dont think it is needed in the gallery. That needs access to all files.

esensar:
> accepting it, but I dont think it is needed in the gallery. That needs access to all files.

Agreed. There should be a way to explicitly ask for all images, or at least to know when user didn't give access to all images to request it again. I will update that gallery PR.

PR #1799, opened 2023-08-23
Feature/manage blocked numbers compose migration

FunkyMuse:
The idea is to test this with TestManageBlockedNumbers activity and then move the code to ManageBlockedNumbersActivity


tibbi:
cant add a number without being the default calling app. Some colors arent changing properly though, like texts and top menu buttons.

FunkyMuse:
> cant add a number without being the default calling app. Some colors arent changing properly though, like texts and top menu buttons.

What do I need to do in order to test this properly, which permissions steps etc.., i've only tested it with fake data?

tibbi:
it is the easiest to test it locally in Simple Dialer or Simple SMS Messenger by including Commons directly, not through jitpack. Just comment out the jitpack commons inclusion and add it locally, steps are in Simple Gallery. At the top of projects gradle file, `implementation project(':commons')` in modules gradle and 2 lines in settings.gradle

commons/src/main/kotlin/com/simplemobiletools/commons/activities/TestManageBlockedNumbers.kt
```diff
@@ -105,6 +106,9 @@ class TestManageBlockedNumbers : BaseSimpleActivity() {
                     },
                     onCopy = { blockedNumber ->
                         copyToClipboard(blockedNumber.number)
+                    },
+                    canScroll = { canPerformScroll ->
+                        canScroll = canPerformScroll
```
tibbi, line 111: isnt this a bit too complicated? We should always assume that this screen can be scrolled, even if no contacts are blocked

FunkyMuse: yeah, it's now a PoC, I don't like the code either, will come up with a better solution, this was just to see if it's working as intended.

tibbi:
so you are rewriting it now or I can accept the PR?

FunkyMuse:
> so you are rewriting it now or I can accept the PR?

i'm rewriting and will begin testing it in the Simple SMS app, will let you know once ready

commons/src/main/kotlin/com/simplemobiletools/commons/compose/screens/ManageBlockedNumbersScreen.kt
```diff
@@ -0,0 +1,528 @@
...
+
+                hasGivenPermissionToBlock && blockedNumbers.isNotEmpty() -> {
+                    items(blockedNumbers, key = { blockedNumber -> blockedNumber.id }) { blockedNumber ->
+                        val isSelected = selectedIds.value.contains(blockedNumber.id)
+                        BlockedNumber(
+                            modifier = Modifier
+                                .animateItemPlacement()
+                                .semantics {
+                                    if (!isInActionMode) {
+                                        onLongClick("select") {
```
tibbi, line 177: whats this "select"?

FunkyMuse: Long press select swipe up/down

tibbi: but what does "select" mean? Is it some command recognized by the system? Shouldnt it be some constant?

FunkyMuse: it's just a label for the `semantics`

commons/src/main/kotlin/com/simplemobiletools/commons/compose/menus/ActionMenu.kt
naveensingh, line 1: Maybe add the gist [link](https://gist.github.com/MachFour/369ebb56a66e2f583ebfb988dda2decf) so the code can be updated from there if need be

tibbi: sure, we can put the link above it, if our code is based on it

tibbi:
lets also limit activity title to 1 line with end ellipsize. If you use max system font size, Frequently asked questions grow to 2 lines. It is good to test something with max system font from time to time.

PR #1800, opened 2023-08-23
feat: try nav color changing

FunkyMuse:
the social and other section are commented to demonstrate the effect

Portrait: non scrollable > transparent nav bar
Landscape: scrollable > primary color nav bar with adjusted alpha

This is PoC to decide, have to update this in other places too


tibbi:
have you tested it by locally including commons? The commented out lines in gallery dont work anymore as some gradle things changed.

FunkyMuse:
> have you tested it by locally including commons? The commented out lines in gallery dont work anymore as some gradle things changed.

This is only for the navigation bar coloring we talked, not about managed blocked numbers

tibbi:
right, the primary color at scrollable navigation looks weird actually, lets use the background color, a bit darkened and translucent

tibbi:
ok, will do more testing in a real app. Thanks

PR #1810, opened 2023-09-02
Make adapter activity variable overridable

naveensingh:
Required for a minor readability change in the music player


FunkyMuse:
> Required for a minor readability change in the music player

Maybe stop depending on the activity and use explicit dependencies or callbacks towards something that is being referenced to and from this class?

naveensingh:
It's mostly used by the adapter itself, not by external components. The adapter code was written years ago, it has a lot of responsibility :)

FunkyMuse:
> It's mostly used by the adapter itself, not by external components. The adapter code was written years ago, it has a lot of responsibility :)

It doesn't make sense to use a tightly coupled activity as a parameter even when used internally, but since it was written years ago it won't matter anyway.

PR #1813, opened 2023-09-04
Keep PhoneNumberConverter class for keeping correct keys

esensar:
This is needed to fix https://github.com/SimpleMobileTools/Simple-Contacts/issues/1026


acathub:
I am suffering the same behaviour as reported with the app crashing at start. Is there a new version available allready?

tibbi:
ye, on Github and it is already In Review on Google Play

acathub:
All contacts are back in great shape thanks 👍

PR #1821, opened 2023-09-09
Add Kurdish language iso 639-1 and iso 639-3 code

GoRaN9O:
Add Kurdish language.

### DIFF
--- a/commons/src/main/res/xml/locale_config.xml
+++ b/commons/src/main/res/xml/locale_config.xml
@@ -7,6 +7,7 @@
     <locale android:name="bn" />
     <locale android:name="br" />
     <locale android:name="ca" />
+    <locale android:name="ckb"/>
     <locale android:name="cs" />
     <locale android:name="cy" />
     <locale android:name="da" />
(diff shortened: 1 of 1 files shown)


tibbi:
why are you adding the language if you havent even translated any strings yet? This file isnt about the keyboard language.

GoRaN909:
> why are you adding the language if you havent even translated any strings yet? This file isnt about the keyboard language.

I added because the keyboard app required to the layout name and I translated the Simple-Keyboard before I add the layout look here is the translated stings of [ckb](https://github.com/SimpleMobileTools/Simple-Keyboard/blob/main/app/src/main/res/values-ckb/strings.xml)

tibbi:
right, this file is about strings though, not the keyboard languages. Will take a look at those soon.

GoRaN909:
> right, this file is about strings though, not the keyboard languages. Will take a look at those soon.

Okay I just started translating some simple tools apps I will do my to completely translate all of them but for now only the keyboard app is important to me :)

PR #1824, opened 2023-09-10
Improve password protection

naveensingh:
### Changes and notes:
 - After 3 incorrect attempts, the user is locked out for 5 seconds. This doesn't distinguish between folder lock vs app lock. 3 incorrect attempts to unlock a folder will cause the whole app to be inaccessible as well. This is because lockin a folder doesn't make sense when app level lock is enabled plus we'll have to complicate things by handling retry count and lock-out time for everything separately.
 - Limit the minimum pin length to 4 digits. This is necessary because even with lockout, there are only 10 combinations one needs to try to break in if the user uses a single-digit PIN. This doesn't affect users already using a PIN shorter than 4 digits.
 -  Move the title to the center and add a lock icon.

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/views/PinTab.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/views/PinTab.kt
@@ -2,31 +2,34 @@ package com.simplemobiletools.commons.views
 
 import android.content.Context
 import android.util.AttributeSet
-import android.view.LayoutInflater
-import android.widget.RelativeLayout
+import android.widget.TextView
 import androidx.biometric.auth.AuthPromptHost
 import com.simplemobiletools.commons.R
 import com.simplemobiletools.commons.databinding.TabPinBinding
 import com.simplemobiletools.commons.extensions.*
 import com.simplemobiletools.commons.helpers.PROTECTION_PIN
+import com.simplemobiletools.commons.interfaces.BaseSecurityTab
 import com.simplemobiletools.commons.interfaces.HashListener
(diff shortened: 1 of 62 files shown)


tibbi:
where did you get the blinking "Wrong patter" lets make it static only. Also do not allow inserting patterns while blocked, or if you cannot do it, at least reset the currenly inserted pattern that is there as the timer goes down to 0. It is now stuck there.

tibbi:
lets also use a bigger icon above Insert patter, it is too tiny

commons/src/main/kotlin/com/simplemobiletools/commons/views/PinTab.kt
```diff
@@ -73,23 +82,38 @@ class PinTab(context: Context, attrs: AttributeSet) : RelativeLayout(context, at
...
+        if (!isLockedOut()) {
+            val newHash = getHashedPin()
+            when {
+                pin.isEmpty() -> {
+                    context.toast(R.string.please_enter_pin)
+                }
+
+                computedHash.isEmpty() && pin.length < MINIMUM_PIN_LENGTH -> {
+                    resetPin()
+                    context.toast(R.string.pin_must_be_4_digits_long)
```
tibbi, line 94: lets show a long toast here

tibbi:
lets not autoaccept the repeated PIN after inserting, let the user click on the check

naveensingh:
okay, applied the suggested changes here: https://github.com/SimpleMobileTools/Simple-Commons/pull/1834

PR #1825, opened 2023-09-10
Feature/add contact blocking helpers

fatihergin:
can be tested on https://github.com/SimpleMobileTools/Simple-Dialer/pull/688

### DIFF
--- a/commons/src/main/res/values-ar/strings.xml
+++ b/commons/src/main/res/values-ar/strings.xml
@@ -139,6 +139,12 @@
     <string name="block_hidden_messages">حظر الرسائل من الأرقام المخفية</string>
     <string name="add_blocked_number_helper_text">أدخل رقما أو نمطا (على سبيل المثال، *12345*، +1*8888) لحظر جميع المكالمات والرسائل من الأرقام المطابقة للنمط.</string>
     <string name="must_make_default_caller_id_app">لا يمكن حظر الأرقام غير المعروفة دون إذن معرف المتصل.</string>
+    <string name="block_contact">Block contact</string>
+    <string name="block_contact_success">Contact blocked</string>
+    <string name="block_contact_fail">Contact could not be blocked</string>
+    <string name="unblock_contact">Unblock contact</string>
+    <string name="unblock_contact_success">Contact unblocked</string>
+    <string name="unblock_contact_fail">Contact could not be unblocked</string>
     <!-- Favorites -->
     <string name="favorites">المفضلة</string>
     <string name="add_favorites">إضافة مفضلة</string>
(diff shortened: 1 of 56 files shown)


tibbi:
seems to work fine, but lets reuse the UI style from Simple Contacts with "Show phone numbers" enabled too, where we arent using bold letters. Contact names should just use full alpha and numbers a lower one. That will make the apps more consistent.

commons/src/main/res/values-ar/strings.xml
```diff
@@ -139,6 +139,12 @@
     <string name="block_hidden_messages">حظر الرسائل من الأرقام المخفية</string>
     <string name="add_blocked_number_helper_text">أدخل رقما أو نمطا (على سبيل المثال، *12345*، +1*8888) لحظر جميع المكالمات والرسائل من الأرقام المطابقة للنمط.</string>
     <string name="must_make_default_caller_id_app">لا يمكن حظر الأرقام غير المعروفة دون إذن معرف المتصل.</string>
+    <string name="block_contact">Block contact</string>
+    <string name="block_contact_success">Contact blocked</string>
+    <string name="block_contact_fail">Contact could not be blocked</string>
```
tibbi, line 144: are we ever using these fails?

tibbi:
> seems to work fine, but lets reuse the UI style from Simple Contacts with "Show phone numbers" enabled too, where we arent ever using bold letters. Contact names should just use full alpha and numbers a lower one. That will make the apps more consistent.

Lets actually ignore this one, we are rewriting some screens including this into Compose anyway, so the way UI is created will change drastically.

commons/src/main/res/values-ar/strings.xml
fatihergin: it was on my enhancement list for the next time, but let me do it now. not a challenging thing

fatihergin: thanks to this commit, we are now able to handle number blocking&unblocking results.
[return contact blocking & unblocking success result](https://github.com/SimpleMobileTools/Simple-Commons/pull/1825/commits/2b77ca6ff191e07370bba9ff703caa5832c730a8)

please check for [sample usage](https://github.com/SimpleMobileTools/Simple-Dialer/pull/688/files#diff-e996a13f4a8a42d520a6e489f1865cbe1fa96b59e3b93e2ec5218163411339f7R188-R199)

fatihergin:
> > seems to work fine, but lets reuse the UI style from Simple Contacts with "Show phone numbers" enabled too, where we arent ever using bold letters. Contact names should just use full alpha and numbers a lower one. That will make the apps more consistent.
> 
> Lets actually ignore this one, we are rewriting some screens including this into Compose anyway, so the way UI is created will change drastically.

well, bold letters were looking so ugly, just removed it without waiting for compose migration :)

tibbi:
the name texts are too small and they are way too to the left side of the screen, not aligned with top checkboxes at all...

tibbi:
okay thanks, this looks good now

PR #1830, opened 2023-09-15
Restore previous logic for selecting layout for WritePermissionDialog

esensar:


tibbi:
what does this change?

esensar:
During the migration to viewbinding, it looks like this was accidentally swapped, which caused dialog for `Mode.OpenDocumentTreeSDK30` and `Mode.CreateDocumentSDK30` to use the wrong layout (to show text about SD card instead of the intended text).

tibbi:
hm ok, thanks

PR #1831, opened 2023-09-15
fix: manage blocked numbers selection and add ability to show names

FunkyMuse:


commons/src/main/kotlin/com/simplemobiletools/commons/compose/screens/ManageBlockedNumbersScreen.kt
```diff
@@ -177,17 +177,21 @@ internal fun ManageBlockedNumbersScreen(
...
-                haptics = LocalHapticFeedback.current,
-                selectedIds = selectedIds,
-                autoScrollSpeed = autoScrollSpeed,
-                autoScrollThreshold = with(LocalDensity.current) { 40.dp.toPx() },
-                dragUpdate = { isDraggingStarted ->
-                    hasDraggingStarted = isDraggingStarted
-                    triggerReset = RESET_IMMEDIATELY
-                }
-            ),
+            modifier = Modifier.ifTrue(!blockedNumbers.isNullOrEmpty()) {
```
tibbi, line 180: this is a really complicated line, cant we simplify it?

FunkyMuse: Sorry I don't understand what's so complicated about it, it checks whether the list is not null or empty

I can create an extension function that reads `isNotNullOrEmpty()` so that it's more verbose, would that clear it up a bit?

tibbi: well, ifTrue with a negation in it. Maybe at least change it to ifFalse and remove the exclamation mark? It already makes it a lot simpler.

tibbi:
ok thanks

PR #1832, opened 2023-09-15
Feature/use correct contact permission while reading contacts

fatihergin:
can be tested with https://github.com/SimpleMobileTools/Simple-SMS-Messenger/pull/748


tibbi:
looks good now, thanks

PR #1834, opened 2023-09-18
Improve password protection further

naveensingh:
### Changes:
 - Removing blinking animation from error text.
 - Disable automatic unlock on the correct PIN.
 - Disable pattern input when the locked-out countdown is in progress (it won't provide any feedback on touch).
 - Increase the dialog's title and icon size. I used 32dp because 40dp and 48dp looked too big.
 - Show long error toasts.


tibbi:
ye looks good, thanks

PR #1842, opened 2023-09-20
feat: add alert dialog state manager

FunkyMuse:


commons/src/main/kotlin/com/simplemobiletools/commons/compose/alert_dialog/AlertDialogState.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    fun show() {
+        isShown = true
+    }
+
+    fun hide() {
+        isShown = false
+    }
+
+    fun toggle() {
```
tibbi, line 40: lets use some clearer name here, like toggleVisibility or so

commons/src/main/kotlin/com/simplemobiletools/commons/compose/alert_dialog/AlertDialogState.kt
```diff
@@ -0,0 +1,56 @@
...
+        isShown = !isShown
+    }
+
+    fun changeValue(predicate: Boolean) {
```
tibbi, line 44: this isnt really clear either

PR #1843, opened 2023-09-20
Fix/otg delete move

esensar:


esensar:
This is required to fix https://github.com/SimpleMobileTools/Simple-File-Manager/issues/602

PR #1844, opened 2023-09-20
feat: migrate add/edit blocked number dialog to Compose

FunkyMuse:


tibbi:
hint text of Add Blocked Number is barely visible in casual light theme, it doesnt change color properly. And in B&W theme the dialog should have pure black background with some white stroke around it, like currently. The rest looks good :)

tibbi:
the app title of App Corrupt uses the primarry color, but I guess it should use the text one, like at other dialogs. And the non-material dialogs are way too wide now, they shouldnt span across while width.

FunkyMuse:
> the dialog title of App Corrupt dialog uses the primary color, but I guess it should use the text one, like at other dialogs. And the non-material dialogs are way too wide now, they shouldnt span across while width.

It used the primary for the title yes as that's how it looked before I migrated it, but I'll change it nevertheless to match other dialogs.

As I understood correctly, the dialogs should use whole width of the screen?

tibbi:
no, dialogs shouldnt use whole width, they should be smaller. The Material You theme dialogs have proper width, the casual old theme ones are too wide.

FunkyMuse:
> no, dialogs shouldnt use whole width, they should be smaller. The Material You theme dialogs have proper width, the casual old theme ones are too wide.

Now i'm using 0.9 width of the screen, maybe 0.7 would make more sense?

tibbi:
check how is material done, try making it consistent

FunkyMuse:
> check how is material dialog done, try making it consistent

Material dialogs by default have it as hardcoded which doesn't make any sense
```kotlin
internal val DialogMinWidth = 280.dp
internal val DialogMaxWidth = 560.dp
```

We can use a screen width percentage to make things persistent, I'll play with the values and see which one looks the best.

tibbi:
width looks good to me now, but the corner should have a much higher radius

tibbi:
it should be even bigger, like the theme picker at Color Customization

tibbi:
Shapes.extraLarge seems to work just fine

samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/MainActivity.kt
```diff
@@ -84,5 +89,7 @@ class MainActivity : BaseSimpleActivity() {
     override fun onResume() {
         super.onResume()
         setupToolbar(binding.mainToolbar)
+
+        PurchaseThankYouDialog(this)
```
tibbi, line 93: remove this, it is quite annoying for testing

tibbi:
this PR can be accepted, right? that test activity isnt an issue to me, but it should be moved into the samples module

FunkyMuse:
> this PR can be accepted, right? that test activity isnt an issue to me, but it should be moved into the samples module

Will move now

tibbi:
it is still large instead of extralarge and the main manifest contains the test activity

PR #1847, opened 2023-09-28
Add helper for creating flows out of config properties

esensar:


tibbi:
what is this for?

esensar:
While working with compose, we will need to access many of these properties in `Config` as Flow. `isBlockingUnknownNumbers` and `isBlockingHiddenNumbers` are examples of these. I added this to make that a bit more simple, so we don't have to repeat the exact same callback code for each property, since it really just wraps the property into a flow.

PR #1852, opened 2023-09-30
feat: add change date time format dialog

FunkyMuse:


commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/ChangeDateTimeFormatDialog.kt
```diff
@@ -71,3 +95,196 @@ class ChangeDateTimeFormatDialog(val activity: Activity, val callback: () -> Uni
...
+        ) {
+            var is24HoursSelected by remember { mutableStateOf(is24HourChecked) }
+
+            val (selected, setSelected) = remember { mutableStateOf(initiallySelected) }
+            RadioGroup(
+                items = kinds, selected = selected,
+                setSelected = setSelected,
+                modifier = Modifier.padding(
+                    vertical = 16.dp,
+                    horizontal = 11.dp
```
tibbi, line 146: lets use radiobuttons that are reusable in other dialogs and apps too, do not hardcode values here

tibbi:
do not use caps lock at February etc

tibbi:
that tiny ripple effect around text itself looks weird, maybe you can add some padding around it?

tibbi:
lets also change the sample year from 2021 to 2023

FunkyMuse:
> lets also change the sample year from 2021 to 2023

February 02, 2023 makes sense?

tibbi:
no, make it february 15 like now, the day is above 12 intentionally to avoid confusion

tibbi:
looks better now, but reduce the margin between radiobuttons to like 2dp line only. And make the ripple stretco to full width to make the top and bottom checkbox consistent

FunkyMuse:
> looks better now, but reduce the margin between radiobuttons to like 2dp line only. And make the ripple stretco to full width to make the top and bottom checkbox consistent

checkbox isn't full width ripple

tibbi:
thats what I mean though, lets make it full width

FunkyMuse:
> thats what I mean though, lets make it full width

Can you check now please

tibbi:
yeah, it is a lot better now, try horizontally aligning the radio circles and bottom checkbox. Also in landscape mode the whole dialog scrolls including Cancel/OK. Shouldnt those buttons be always visible and only the content should scroll? Thats how it worked previously and I prefer it, always seeing Cancel/OK.

FunkyMuse:
> yeah, it is a lot better now, try horizontally aligning the radio circles and bottom checkbox. Also in landscape mode the whole dialog scrolls including Cancel/OK. Shouldnt those buttons be always visible and only the content should scroll? Thats how it worked previously and I prefer it, always seeing Cancel/OK.

Done, also added star rating dialog

tibbi:
yeah, looks great. One more details, can we show the scrollbar for a moment on the side, if content actually scrolls? It happened in the past that in the gallery sorting dialog people havent noticed that there is an additional checkbox at the bottom, because they had no idea that it actually scrolls. Maybe this small indication would help a bit.

tibbi:
as per discussion, compose doesnt support it properly yet, so accepting it

PR #1855, opened 2023-10-03
feat: upgrade to pro dialog and radio group dialog

FunkyMuse:
Upgrade to pro is needed for simple thank you


FunkyMuse:
@tibbi this PR is the last one before we migrate "Thank you" fully to Compose with the Dialogs

FunkyMuse:
> the top and bottom spacings at Change view type dialog are way too big. Also clicking on one shouldnt close the dialog as there is usually a checkbox at the bottom too. And it should be possible to dismiss it by clicking outside of it. Ill do more testing of those in a real app.

Done

FunkyMuse:
> not really, clicking outside the dialog still doesnt close Change view type. And make the ripple span across whole width to make dialogs consistent. And isnt the font smaller than at Radio Group dialog?

Done, no, font is the same

tibbi:
yeah thanks, looks good now

PR #1856, opened 2023-10-03
Add reusable dimension for compose

esensar:


esensar:
I would prefer to move dimensions from `Dimens.text` to be part of `MaterialTheme.typography`, but I was not sure of mapping between these styles and our sizes, so I decided for this approach for now.

If there is a general idea for a couple of styles that we always use throughout the application, we could put these into typography for easier reuse.

commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/model/Dimensions.kt
```diff
@@ -0,0 +1,42 @@
+package com.simplemobiletools.commons.compose.theme.model
+
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.TextUnit
+
+data class Dimensions(
```
FunkyMuse, line 6: Mark all classes as `@Stable`

FunkyMuse (changes requested):
we need to make sure the classes are stable to avoid shadow recomposition

FunkyMuse:
> I would prefer to move dimensions from `Dimens.text` to be part of `MaterialTheme.typography`, but I was not sure of mapping between these styles and our sizes, so I decided for this approach for now.
> 
> If there is a general idea for a couple of styles that we always use throughout the application, we could put these into typography for easier reuse.

The typography thing is for providing a `TextStyle`

esensar:
> > I would prefer to move dimensions from `Dimens.text` to be part of `MaterialTheme.typography`, but I was not sure of mapping between these styles and our sizes, so I decided for this approach for now.
> > If there is a general idea for a couple of styles that we always use throughout the application, we could put these into typography for easier reuse.
> 
> The typography thing is for providing a `TextStyle`

Yeah, but I figure there are many styles that are reused throughout the app, which also would define a text size, but I haven't really been able to properly figure out which ones to map from the `styles.xml` that we currently have.

FunkyMuse:
> > > I would prefer to move dimensions from `Dimens.text` to be part of `MaterialTheme.typography`, but I was not sure of mapping between these styles and our sizes, so I decided for this approach for now.
> > > If there is a general idea for a couple of styles that we always use throughout the application, we could put these into typography for easier reuse.
> > 
> > 
> > The typography thing is for providing a `TextStyle`
> 
> Yeah, but I figure there are many styles that are reused throughout the app, which also would define a text size, but I haven't really been able to properly figure out which ones to map from the `styles.xml` that we currently have.

you need to map everything inside the `Typography` class which is 

```kotlin
 val displayLarge: TextStyle = TypographyTokens.DisplayLarge,
    val displayMedium: TextStyle = TypographyTokens.DisplayMedium,
    val displaySmall: TextStyle = TypographyTokens.DisplaySmall,
    val headlineLarge: TextStyle = TypographyTokens.HeadlineLarge,
    val headlineMedium: TextStyle = TypographyTokens.HeadlineMedium,
    val headlineSmall: TextStyle = TypographyTokens.HeadlineSmall,
    val titleLarge: TextStyle = TypographyTokens.TitleLarge,
    val titleMedium: TextStyle = TypographyTokens.TitleMedium,
    val titleSmall: TextStyle = TypographyTokens.TitleSmall,
    val bodyLarge: TextStyle = TypographyTokens.BodyLarge,
    val bodyMedium: TextStyle = TypographyTokens.BodyMedium,
    val bodySmall: TextStyle = TypographyTokens.BodySmall,
    val labelLarge: TextStyle = TypographyTokens.LabelLarge,
    val labelMedium: TextStyle = TypographyTokens.LabelMedium,
    val labelSmall: TextStyle = TypographyTokens.LabelSmall,
```

esensar:
Yeah, that is the reason I gave up on that for this PR, since it is not really clear how we want to set up each of these.

FunkyMuse:
I think it's better to copy the existing M3 tokens already and just do a slight modification to the sizes etc...

esensar:
> I think it's better to copy the existing M3 tokens already and just do a slight modification to the sizes...
> 
> Either way a typography style is a data class, you can copy it and change other params.

Yeah, agreed, but still it would be good to give it a bit more time, since it would be good to make sure these make sense before using them throughout the codebase, especially since some of these will probably stay at default values.

esensar:
But, either way, I think this stuff from the current PR is alright to bridge the gap between the `XML` dimens. I think we need to alter these ones too, or at least try to figure out a bit better naming, since some of these are confusing (e.g. `smaller` being bigger than `small`).

FunkyMuse:
We can continue with this approach until we start using Compose in more than one app and come back to re-do the `Typography` that way we can have a grasp of what needs and how to be done.

esensar:
Yeah, I guess the `Simple-Flashlight` was not a really good fit for that task, considering it has almost no text at all.

FunkyMuse:
Come to thinks about it, the reason why I didn't do this in a similar way is because hardcoded values are easily replaceable but the wrong approach is kinda hard to refactor, maybe we should leave them hardcoded until we see how to proceed when working on an app that has texts.

esensar:
> Come to thinks about it, the reason why I didn't do this in a similar way is because hardcoded values are easily replaceable but the wrong approach is kinda hard to refactor, maybe we should leave them hardcoded until we see how to proceed when working on an app that has texts.

Alright, I could remove the `text` from `Dimensions`. I think others are worth keeping, but it would maybe be good to sort out the naming right away, so we don't have to change it once it is spread to multiple apps.

FunkyMuse:
> > Come to thinks about it, the reason why I didn't do this in a similar way is because hardcoded values are easily replaceable but the wrong approach is kinda hard to refactor, maybe we should leave them hardcoded until we see how to proceed when working on an app that has texts.
> 
> Alright, I could remove the `text` from `Dimensions`. I think others are worth keeping, but it would maybe be good to sort out the naming right away, so we don't have to change it once it is spread to multiple apps.
> 
> Or do you think we should avoid this completely until we get some standards in place?

I think we should avoid this completely until we get some standards in place or at least get a grasp of things in the apps, we shouldn't jump to conclusions from only one app that has barely any text in it.

It will be a bit messy at first with hardcoded values but IMO in the long run with a better approach we'll have longer mileage or something like that.

esensar:
Alright, makes sense, I think we can close this for now.

FunkyMuse:
> > Come to thinks about it, the reason why I didn't do this in a similar way is because hardcoded values are easily replaceable but the wrong approach is kinda hard to refactor, maybe we should leave them hardcoded until we see how to proceed when working on an app that has texts.
> 
> Alright, I could remove the `text` from `Dimensions`. I think others are worth keeping, but it would maybe be good to sort out the naming right away, so we don't have to change it once it is spread to multiple apps.
> 
> Or do you think we should avoid this completely until we get some standards in place?

Only for the text sorry, icons sizes everything else can remain but with better naming @esensar

esensar:
Okay, I will see if I can make something more sensible and maybe eliminate some less often used dimensions from that list.

FunkyMuse:
> Okay, I will see if I can make something more sensible and maybe eliminate some less often used dimensions from that list.

We can draw inspiration from Material naming for the sizing for starters, we're using material after all

esensar:
> > Okay, I will see if I can make something more sensible and maybe eliminate some less often used dimensions from that list.
> 
> We can draw inspiration from Material naming for the sizing for starters, we're using material after all

Do you mean the naming that is used in `Shapes`?

esensar:
By the way, since we are renaming stuff, I think it would be good to rename `margin` to `spacing`, since `margin` is not really used throughout `Compose` and `spacing` would mean something more generic, since we would use it for `padding` but also empty spaces in some cases too.

esensar:
And also I think it makes sense to limit number of shared dimension values, since in many cases, these should be reused by reusing components which would include these specific values.

FunkyMuse:
> > > Okay, I will see if I can make something more sensible and maybe eliminate some less often used dimensions from that list.
> > 
> > 
> > We can draw inspiration from Material naming for the sizing for starters, we're using material after all
> 
> Do you mean the naming that is used in `Shapes`?

yes

FunkyMuse:
> By the way, since we are renaming stuff, I think it would be good to rename `margin` to `spacing`, since `margin` is not really used throughout `Compose` and `spacing` would mean something more generic, since we would use it for `padding` but also empty spaces in some cases too.

Padding is the way to go for Compose

FunkyMuse:
> ense to reuse outside of these contexts, which could be encap

Correct, the dimens would definitely live only inside the component, by creating and utilizing same components everywhere, the apps would look consistent.

esensar:
So based on the number of usages, I think we can reduce to these 5 sizes:
```
extraLarge = activity_margin = 16dp
large = normal_margin = 12dp
medium = medium_margin = 8dp
small = small_margin = 4dp
extraSmall = tiny_margin = 2dp
```

FunkyMuse:
> So based on the number of usages, I think we can reduce to these 5 sizes:
> 
> ```
> extraLarge = activity_margin = 16dp
> large = normal_margin = 12dp
> medium = medium_margin = 8dp
> small = small_margin = 4dp
> extraSmall = tiny_margin = 2dp
> ```

Looks good to me

tibbi:
so is this PR ready to be merged or you are still tweaking something?

commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/Dimensions.kt
```diff
@@ -0,0 +1,33 @@
+package com.simplemobiletools.commons.compose.theme
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.ProvidableCompositionLocal
+import androidx.compose.runtime.ReadOnlyComposable
+import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.compose.ui.unit.dp
+import com.simplemobiletools.commons.compose.theme.model.Dimensions
+
+val Dimens: Dimensions
```
FunkyMuse, line 10: Don't create this type of shortcuts, it's a double edged sword, it's better to use `val dimens = LocalDimensions.current` IMO

commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/Theme.kt
```diff
@@ -65,13 +67,17 @@ internal fun Theme(
         updateRecentsAppIcon(baseConfig, context)
     }
 
+    val dimensions = if (configuration.screenWidthDp <= 600) CommonDimensions else Sw600DpDimensions
```
FunkyMuse, line 70: It's better to approach it with https://developer.android.com/reference/kotlin/androidx/compose/material3/windowsizeclass/WindowSizeClass

esensar: Yeah, not sure why I did this. I wanted to keep it the same as previously, but I guess anything other than compact would be the same as this. Anyways, I think I will remove that other class for now, since it is exactly the same as `CommonDimensions` for now.

commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/Dimensions.kt
esensar: To be honest, I would have liked something like `MaterialTheme` object, that encapsulates a couple of properties that act like this, making it much harder to accidentally import these shortcuts accidentally, but I was not sure how to name it so I left it like this for now. I can change to `dimens`, unless you have a suggestion for an object name that would hold this and maybe some other similar properties in the future.

esensar:
> so is this PR ready to be merged or you are still tweaking something? Havent tested it yet.

Nothing to test here, just defining some common properties (similar to what we had in `XML` resources previously, but for compose). If @FunkyMuse approves, I think it is ready to merge.

FunkyMuse:
> > so is this PR ready to be merged or you are still tweaking something? Havent tested it yet.
> 
> Nothing to test here, just defining some common properties (similar to what we had in `XML` resources previously, but for compose). If @FunkyMuse approves, I think it is ready to merge.

I'm about totally removing the 

```
val dimens: Dimensions
    @Composable
    @ReadOnlyComposable
    get() = LocalDimensions.current
```

`Dimens` was better than `dimens`, to me this is a shortcut, I think we can proceed just revert it back to a first capital and if later on we see this as a hurdle when coding we would start using `LocalDimensions.current` everywhere, just do the renaming and we're ready.

esensar:
Reverted. Still I am hoping to find a nicer solution, but I think this is alright for now.

FunkyMuse:
This is the way to go, `MaterialTheme` does it like this as well, maybe we can have something like

```kotlin
object SimpleTheme {

    val dimens: Dimensions
        @Composable
        @ReadOnlyComposable
        get() = LocalDimensions.current 
}
```

later on when we have more things to add, we'll just throw them there that are related, what do you think?

esensar:
Yeah, I guess that is the way to go, since it won't pollute global namespace. Hopefully it won't be confusing, having both `MaterialTheme` and `SimpleTheme`.

FunkyMuse:
> Yeah, I guess that is the way to go, since it won't pollute global namespace. Hopefully it won't be confusing, having both `MaterialTheme` and `SimpleTheme`.

We can merge the `MaterialTheme` values into `SimpleTheme`

FunkyMuse:
we can merge this @tibbi

tibbi:
alright thanks

PR #1858, opened 2023-10-04
Add compose version of ColorPickerDialog

esensar:


esensar:
In the end I had to fork the library, because it didn't allow changing brightness from the outside, meaning that manually entering hex codes would result in different colors, unless they had the exact same brightness as the one that is currently selected.

commons/src/main/res/values-bs/strings.xml
```diff
@@ -0,0 +1,1205 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tibbi, line 1: you translated this file?

esensar: Yes, I mean most of the strings are the same as in Croatian (`values-hr`), with some minor differences.

commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/ColorPickerDialog.kt
```diff
@@ -248,8 +272,263 @@ class ColorPickerDialog(
...
     private fun getVal() = currentColorHsv[2]
 }
+
+@Composable
+fun ColorPickerAlertDialog(
+    alertDialogState: AlertDialogState,
+    modifier: Modifier = Modifier,
+    @ColorInt
+    color: Int,
+    recentColors: List<Int>,
```
FunkyMuse, line 287: This needs to be an `ImmutableList` to avoid unnecessary recompositions

commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/ColorPickerDialog.kt
```diff
@@ -248,8 +272,263 @@ class ColorPickerDialog(
...
+
+    AlertDialog(
+        modifier = modifier
+            .dialogBorder,
+        onDismissRequest = alertDialogState::hide,
+        properties = DialogProperties(usePlatformDefaultWidth = false)
+    ) {
+        Column(
+            modifier = Modifier
+                .fillMaxWidth(0.95f)
```
FunkyMuse, line 305: This doesn't work on Samsung devices :(

https://issuetracker.google.com/issues/260755409

esensar: Oh, I just copied that from `LineColorPickerDialog` to ensure same behavior. I will take a look.

FunkyMuse: I can test it on my S22 U, will attach screenshot in a bit.

esensar: Great, thanks!

FunkyMuse: It seems to be fixed using OneUI +5.0

commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/ColorPickerDialog.kt
```diff
@@ -248,8 +272,263 @@ class ColorPickerDialog(
...
+                .background(dialogContainerColor, dialogShape)
```
FunkyMuse, line 306: you can use `.dialogBackgroundAndShape`

commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/ColorPickerDialog.kt
esensar: Yeah, I saw that in the thread you sent. Is this okay to keep then?

FunkyMuse: It's best if @tibbi tests it too on his Samsung device, if it works let's keep it, this is a good sign.

commons/src/main/kotlin/com/simplemobiletools/commons/dialogs/ColorPickerDialog.kt
FunkyMuse: We always take only **5** of the recent colors so....

tibbi:
so the dialog in customization activity is rewritten already? It looks exactly the same as the old one?

esensar:
> Im not a fan of the new color picker, that bottom bar is not practical and the whole dialog is unnecessarily huge. Id say lets keep the old one for now.

Alright, I can wrap it in compose, similar to `LineColorPickerDialog`

tibbi:
yeah, lets do that

tibbi:
so thanks, but Ill close this

PR #1859, opened 2023-10-04
feat: About activity compose dialogs and extensions

FunkyMuse:
- add extensions for showing app launched dialogs
- migrate About activity to use Compose dialogs

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/compose/extensions/AcitivtyExtensions.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/compose/extensions/AcitivtyExtensions.kt
@@ -1,0 +1,101 @@
+package com.simplemobiletools.commons.compose.extensions
+
+import android.content.ComponentName
+import android.content.Context
+import android.content.pm.PackageManager
+import androidx.activity.ComponentActivity
+import androidx.core.content.ContextCompat
+import com.simplemobiletools.commons.R
+import com.simplemobiletools.commons.extensions.*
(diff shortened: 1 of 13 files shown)


commons/src/main/kotlin/com/simplemobiletools/commons/compose/extensions/AcitivtyExtensions.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    baseConfig.lastVersion = currVersion
+}
+
+fun ComponentActivity.upgradeToPro() {
+    launchViewIntent("https://simplemobiletools.com/upgrade_to_pro")
+}
+
+const val DEV_PS_URL = "https://play.google.com/store/apps/dev?id=9070296388022589266"
```
tibbi, line 89: whats PS? This isnt  a clear constant at all... Make it DEVELOPER_URL or so

tibbi:
when the use clicks on "Rate us" on About page, then Skip, we should show the RateStars dialog, not redirect directly.

FunkyMuse:
> when the use clicks on "Rate us" on About page, then Skip, we should show the RateStars dialog, not redirect directly. Redirect the user only if he picks 5 stars.

Done

tibbi:
ok thanks, should be good. Will do more testing in a real app.

FunkyMuse:
Will merge this into thank you

sudoshindo:
@tibbi sellout

PR #1868, opened 2023-10-05
Add compose version of `CheckAppOnSdCard`

esensar:


commons/src/main/kotlin/com/simplemobiletools/commons/compose/extensions/AcitivtyExtensions.kt
```diff
@@ -99,3 +100,12 @@ fun Context.fakeVersionCheck(
         }
     }
 }
+
+fun Activity.appOnSdCardCheck(
```
FunkyMuse, line 104: For compose this should be preferably `ComponentActivity` so that we can distinguish, also it should be named better so that we know as well `appOnSdCardCheckCompose` later on when migrate all the apps we'll just remove the suffix

FunkyMuse (approved):

PR #1869, opened 2023-10-05
Add separate `AppTheme` composable without any surface

esensar:
This is needed for screens like widget configuration, which have fully transparent backgrounds

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/AppTheme.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/AppTheme.kt
@@ -37,6 +37,7 @@ fun AppTheme(
     TransparentSystemBars()
     Theme(theme = currentTheme) {
         content()
+        OnContentDisplayed()
     }
 }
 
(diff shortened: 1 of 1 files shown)


FunkyMuse (approved):

commons/src/main/kotlin/com/simplemobiletools/commons/compose/theme/AppTheme.kt
```diff
@@ -14,8 +14,7 @@ import com.simplemobiletools.commons.compose.theme.model.Theme
 import com.simplemobiletools.commons.compose.theme.model.Theme.Companion.systemDefaultMaterialYou
 
 @Composable
-fun AppThemeSurface(
-    modifier: Modifier = Modifier,
+fun AppTheme(
```
FunkyMuse, line 17: Keep in mind, we don't have the check here with `OnContentDisplayed`

esensar: Yeah, I was wondering if I should put it there. It shouldn't be an issue that it is outside of the `Surface` right? Since it only displays a `Dialog` which should have its own?

FunkyMuse: The dialogs we have are already handling their background colors etc..

I don't think I'll be an issue, but please test it, I'm finishing up for today, started early gotta leave early, will see the other PR tomorrow when it's done

PR #1877, opened 2023-10-09
Fix colors for dropdown menus

esensar:


esensar:
I am not sure why these dropdown menus don't pick up theme values by themselves. They should behave similarly to dialogs, but I am not sure what I am missing here. It worked after I manually set values in modifier and `Text` composable.

tibbi:
menu colors seem to be good, but the top status and bottom navigation colors seem to be broken on the main screen. Just recheck how should it looks like, both material and non-material themes are wrong.

gradle.properties
```diff
@@ -16,3 +16,4 @@ android.useAndroidX=true
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 org.gradle.jvmargs=-Xmx8192m
+android.nonTransitiveRClass=false
```
FunkyMuse, line 19: Nope, remove this, this is true by default now

FunkyMuse:
why are we migrating the main screen in the samples to Compose, is it necessary?

I think we should leave it for backwards compatibility to check how things "were" and build upon that.

esensar:
> why are we migrating the main screen in the samples to Compose, is it necessary?
> 
> I think we should leave it for backwards compatibility to check how things "were" and build upon that.

I migrated it for easier testing, because I guess compose is the way to go for all the apps. For checking for how things were, I thought we would use the apps that we are migrating to Compose and not main screen from `Commons`.

gradle.properties
esensar: My bad, I thought I should use `Simple-Thank-You` as reference and it has it set to false.

samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/MainActivity.kt
```diff
@@ -2,55 +2,178 @@ package com.simplemobiletools.commons.samples.activities
...
         super.onCreate(savedInstanceState)
-        setContentView(binding.root)
-        appLaunched(BuildConfig.APPLICATION_ID)
 
-        updateMaterialActivityViews(binding.mainCoordinator, binding.mainHolder, useTransparentNavigation = true, useTopSearchMenu = false)
-        setupMaterialScrollListener(binding.mainNestedScrollview, binding.mainToolbar)
+        enableEdgeToEdgeSimple()
+        setContent {
+            AppThemeSurface {
+                SettingsLazyScaffold(
```
FunkyMuse, line 55: update to latest commons please, the name has been changed

samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/MainActivity.kt
```diff
@@ -2,55 +2,178 @@ package com.simplemobiletools.commons.samples.activities
...
+                                    listOf(about, moreApps).toImmutableList()
+                                }
+                                var isMenuVisible by remember { mutableStateOf(false) }
+                                ActionMenu(
+                                    items = actionMenus,
+                                    numIcons = 2,
+                                    isMenuVisible = isMenuVisible,
+                                    onMenuToggle = { isMenuVisible = it },
+                                    iconsColor = scrolledColor
+                                )
```
FunkyMuse, line 82: Please extract in a function for better readability

samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/MainActivity.kt
FunkyMuse: https://github.com/SimpleMobileTools/Simple-Commons/pull/1879

samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/MainActivity.kt
```diff
@@ -52,7 +52,7 @@ class MainActivity : ComponentActivity() {
         enableEdgeToEdgeSimple()
         setContent {
             AppThemeSurface {
-                SettingsLazyScaffold(
+                SimpleLazyListScaffold(
```
FunkyMuse, line 55: it will change to SimpleScaffold soon with the new PR FYI as i think it makes more sense

samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/MainActivity.kt
esensar: I will also extract the screen then, to make it a bit more similar to how it should be done in actual apps.

samples/src/main/kotlin/com/simplemobiletools/commons/samples/activities/MainActivity.kt
esensar: Updated.

FunkyMuse:
I'll review it today as early as possible as I can

commons/src/main/kotlin/com/simplemobiletools/commons/compose/menus/ActionMenu.kt
```diff
@@ -140,6 +142,9 @@ fun ActionMenu(
             }
         }
         DropdownMenu(
+            modifier = Modifier
+                .background(SimpleTheme.colorScheme.surface)
```
FunkyMuse, line 146: this should be the same as `dialogContainerColor`

FunkyMuse (approved):
Please change the background color and we're good to go, I have a plane to catch, once you do that change @tibbi please merge it.

FunkyMuse (approved):

FunkyMuse:
@tibbi we can merge this now

PR #1889, opened 2023-10-13
Add `sendOnCollect` flag to `sharedPreferencesCallback` for consistent values flow

esensar:
@FunkyMuse this is related to https://github.com/SimpleMobileTools/Simple-Flashlight/pull/213

I tested it in this sample app by making changes on `TestDialogActivity` and checking if they are reflected on `MainActivity`. They weren't reflected previously, but with this approach they are (only when `sendOnCollect` is true of course).

### DIFF
--- a/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/SharedPreferencesProducerExtensions.kt
+++ b/commons/src/main/kotlin/com/simplemobiletools/commons/extensions/SharedPreferencesProducerExtensions.kt
@@ -7,6 +7,7 @@ import kotlinx.coroutines.flow.callbackFlow
 
 context (SharedPreferences)
 fun <T> sharedPreferencesCallback(
+    sendOnCollect: Boolean = false,
     value: () -> T?,
 ): Flow<T?> = callbackFlow {
     val sharedPreferencesListener =
@@ -14,6 +15,9 @@ fun <T> sharedPreferencesCallback(
             trySend(value())
         }
 
+    if (sendOnCollect) {
+        trySend(value())
+    }
     registerOnSharedPreferenceChangeListener(sharedPreferencesListener)
     awaitClose { unregisterOnSharedPreferenceChangeListener(sharedPreferencesListener) }
 }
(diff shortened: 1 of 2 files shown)


FunkyMuse:
Can you add a sample code, this seems more of a workaround than a solution

esensar:
> Can you add a sample code in the Flashlight migration, this seems more of a workaround than a solution

How do you mean? Sample of what? If you mean this, then the change in that migration would just be setting this flag to true.

On the other hand, I understand this looking more like a workaround. This should either be the default behavior or we can remove it completely, but then look for a different solution for cases when changes to preferences are done outside of observer lifecycle.

FunkyMuse:
Or we can also send the value to the callback after we register the listener, that could work? 🤷‍♂️

esensar:
> Or we can also send the value to the callback after we register the listener, that could work? 🤷‍♂️

I am not sure I am following... How is that different from this?

FunkyMuse:
> > Or we can also send the value to the callback after we register the listener, that could work? 🤷‍♂️
> 
> I am not sure I am following... How is that different from this?

We're not emitting the initial value, only when it changes, I'm just thinking out loud

esensar:
> > > Or we can also send the value to the callback after we register the listener, that could work? 🤷‍♂️
> > 
> > 
> > I am not sure I am following... How is that different from this?
> 
> We're not emitting the initial value, only when it changes, I'm just thinking out loud

Yeah, that is what removing this flag and just making this the default behavior would do. But on the other hand, I think this function currently does exactly as it says - it publishes changes to a specific preference key as long as you observe it.

Now, we can either patch it up so it works considering the nature of lifecycles, or add a helper as an additional layer on top of it (like you suggested with state flow).

FunkyMuse:
> > > > Or we can also send the value to the callback after we register the listener, that could work? 🤷‍♂️
> > > 
> > > 
> > > I am not sure I am following... How is that different from this?
> > 
> > 
> > We're not emitting the initial value, only when it changes, I'm just thinking out loud
> 
> Yeah, that is what removing this flag and just making this the default behavior would do. But on the other hand, I think this function currently does exactly as it says - it publishes changes to a specific preference key as long as you observe it.
> 
> Now, we can either patch it up so it works considering the nature of lifecycles, or add a helper as an additional layer on top of it (like you suggested with state flow).

Let's have the two, with better naming of course.

esensar:
Also, just thinking about that state flow thing, if I understood what you meant, it means we would always be observing preferences, regardless of any uses of these values?

esensar:
> I think we can unregister the listener when the flow completes `invokeOnCompletion` but please test it

Wouldn't that result in same behavior?

FunkyMuse:
> > I think we can unregister the listener when the flow completes `invokeOnCompletion` but please test it
> 
> Wouldn't that result in same behavior?

we can proceed with this workaround if you're sure it works

FunkyMuse (approved):

PR #1892, opened 2023-10-17
Add compose version of `FilePickerDialog`

esensar:


esensar:
Similar to other dialogs, it does not act exactly the same as original dialogs, since some of the handling needs to be done outside of the dialog.

tibbi:
the UI sadly looks way too bad and there are many breaking changes + magic numbers, lets avoid rewriting this dialog. Thanks

